PR #7, opened 2021-05-28
add back __len__ function for MDASequence 

fdrgsp:
add back __len__ function for MDASequence


tlambert03:
wanna comment on what the issue was?

fdrgsp:
Nothing, I wanted to use the np.prod of the length of the sequence as a way to save the final mad stack. But I found a better way using the shape of the sequence.

PR #35, opened 2022-03-05
add MDAsequence level metadata

ianhi:
Seems natural to have metadata at the sequence level rather than just the event level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ multidimensional acquisition GUI in most microscope software).  A
 # simplified, and possibly outdated.  See useq.MDASequence in codebase
 
 class MDASequence(BaseModel):
+    metadata: Dict[str, Any] = {}  # user-specific data
     axis_order: str  # e.g. 'tpcz'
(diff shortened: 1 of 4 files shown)


tlambert03:
yeah, this is great.  One slight concern though is serialization.  This is definitely an object that should be serializeable.  If a user adds a complex python object here, it will in some cases break that.  Of course that will only affect that user, and will only occur when they're using useq in a context that actually does require serialization (such as a remote CMMCore, or over ZMQ for example) ... but we might want to consider how that would be handled, or whether we also now need to let the user declare how objects should be serialized

ianhi:
> yeah, this is great. One slight concern though is serialization. This is definitely an object that should be serializeable.

hmm indeed. Though important to note that this PR doesn't introduce that issue as it already existed on the MDAEvent.  did we just never run into that limit when using remote, or were the pyro serializers better at handling complex python objects than json.dumps?

tlambert03 (approved):
thanks!

PR #42, opened 2022-05-24
fix: fix z_pos when there is no z_plan

fdrgsp:
At the moment, when there is _**no**_ `z_plan` (`NoZ`), when we set the stage to move and acquire images at different xy positions (e.g. through the `napari-micromanager` `mda` position table), the `z_pos` is always `None` even if it is defined in the experiment.

This PR fix this problem in the `iter_events` methods of the `MDASequence` class.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,11 +10,11 @@ repos:
       - id: autoflake
         args: ["--in-place", "--remove-all-unused-imports"]
   - repo: https://github.com/PyCQA/isort
-    rev: 5.9.2
+    rev: 5.9.3
     hooks:
       - id: isort
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.23.0
+    rev: v2.23.1
     hooks:
       - id: pyupgrade
         args: ["--py3-plus", "--keep-runtime-typing"]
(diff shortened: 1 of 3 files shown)


tlambert03 (commented):
is this ready to go?  Is there a test you could add that fails on main for 

> when we set the stage to move and acquire images at different xy positions (e.g. through the napari-micromanager mda position table), the z_pos is always None even if it is defined in the experiment.

tlambert03:
> > is this ready to go? Is there a test you could add that fails on main for
> 
> yes it is ready to go. There is no error (it works), it is only not considering the z position if there is no z_plan. 

well ... you said:

> when we set the stage to move and acquire images at different xy positions (e.g. through the napari-micromanager mda position table), the z_pos is always None even if it is defined in the experiment.

and then you opened this PR to fix it ... so, looking at the code you submitted, it looks to me like creating a MDASequence and then calling `iter_events()` should have different behavior now than before right?  that's what I mean... can you add a test for that?

fdrgsp:
@tlambert03 

you mean something simple like this?

```
def test_z_position() -> None:
    mda = MDASequence(
        axis_order="tpcz",
        stage_positions=[(222, 1, 10), (111, 1, 20)]
    )
    for event in mda:
        assert not mda.z_plan
        assert event.z_pos
```

tlambert03:
yep, exactly.  Something that fails on the `main` branch, but is fixed here in your PR.  something to make sure that whatever you've done here won't be "lost" accidentally in the future.  tests can be super simple

PR #44, opened 2022-06-27
feat: Expose position name in MDAEvent

fdrgsp:
This PR expose in the`MDAEvent` the position `name` stored in the `Position` class used in `MDASequence`.

Used in https://github.com/tlambert03/napari-micromanager/pull/180.


tlambert03:
Lemme know when this is not wip

fdrgsp:
> Lemme know when this is not wip

@tlambert03 not WIP :)

fdrgsp:
@tlambert03 I updated also the tests. Now it is not wip anymore.

useq/_position.py
```diff
@@ -29,6 +29,20 @@ def validate(cls, value: Any) -> Position:
         if isinstance(value, (np.ndarray, tuple)):
             x, *value = value
             y, *value = value or (None,)
-            z = value[0] if value else None
-            return Position(x=x, y=y, z=z)
+            if value:
```
tlambert03, line 32: this is pretty hard to visually parse and reason about.  Can you add some comments to show examples of the cases that led you to add all these conditionals?

tlambert03: moreover... do we _have_ to support the case of passing name in a tuple?  Can we not just state that, if you'd like to pass an object with a name, it has to be a `dict`?

fdrgsp: @tlambert03 

> this is pretty hard to visually parse and reason about. Can you add some comments to show examples of the cases that led you to add all these conditionals?

The conditionals are checking if the `value` contains `z`, `name` or both info:

(100, 200, 300)  -> x = 100, y = 200, z = 300
(100, 200, "pos_name") -> x = 100, y = 200, name ="pos_name"
(100, 200, 300, "pos_name") -> x = 100, y = 200, z = 300, name ="pos_name"

> moreover... do we have to support the case of passing name in a tuple? Can we not just state that, if you'd like to pass an object with a name, it has to be a dict?

Sure, I thought it was good to make it available also when using tuples but It is definitely not necessary. Should I remove it?

tlambert03: yeah, "yagni" :)
a 3-tuple of integers for XYZ makes sense, but once we start adding other types that don't naturally fit as an array, let's just not accept it as a tuple

tlambert03 (approved):
nice and simple. LGTM, thanks!

PR #70, opened 2023-01-22
feat: add tile dimension

fdrgsp:
This PR adds the tile dimension to enable grid acquisition experiments.

TODO:
- `TileFromCorners`: find how to define the `top_left` starting position.
- find a way to discriminate between a "general" `tile_plan` and a `tile_plan` specified in the `Position` object. (NOTE: the `Position` class allows to also set the z_plan but we are not accessing it at the moment, we should fix this as well).
- update tests.


tests/test_json.py
```diff
@@ -2,8 +2,7 @@
...
-
+JSON = '{"metadata": {"some info": "something"}, "axis_order": "tpcz", "stage_positions": [{"x": 10.0, "y": 20.0, "z": null, "name": null, "z_plan": {"go_up": true}, "tile_plan": {"overlap": 0.0, "snake_order": true}}, {"x": 10.0, "y": 20.0, "z": 50.0, "name": "test_name", "z_plan": {"go_up": true, "above": 10.0, "below": 0.0, "step": 1.0}, "tile_plan": {"overlap": 0.0, "snake_order": true}}], "tile_plan": {"overlap": 0.0, "snake_order": true, "rows": 2, "cols": 1, "relative_to": "center"}, "channels": [{"config": "Cy5", "group": "Channel", "exposure": 50.0, "do_stack": true, "z_offset": 0.0, "acquire_every": 1, "camera": null}, {"config": "FITC", "group": "Channel", "exposure": 100.0, "do_stack": true, "z_offset": 0.0, "acquire_every": 1, "camera": null}, {"config": "DAPI", "group": "Channel", "exposure": null, "do_stack": false, "z_offset": 0.0, "acquire_every": 3, "camera": null}], "time_plan": {"prioritize_duration": false, "phases": [{"prioritize_duration": false, "interval": 3.0, "loops": 3}, {"prioritize_duration": true, "interval": 10.0, "duration": 2400.0}]}, "z_plan": {"go_up": true, "range": 1.0, "step": 0.5}}'  # noqa
```
tlambert03, line 5: let's put this into a new nicely formatted .json file (and load it as a fixture) ... makes it much better for git to see what has changed (I should have done that in the first place)

tlambert03: actually never mind, I'll do that in another pr

tlambert03: ok, you can merge in main now

src/useq/_mda_sequence.py
```diff
@@ -63,23 +67,27 @@ class MDASequence(UseqModel):
...
+    Multi-Dimensional Acquisition ▶ nt: 2, np: 1, nc: 1, nz: 4, ng: 4
 
     >>> for event in seq:
     ...     print(event)
 
     >>> print(seq.yaml())
     channels:
     - config: DAPI
-      exposure: 1.0
+        exposure: 1.0
```
tlambert03, line 83:
```suggestion
      exposure: 1.0
```

src/useq/_mda_sequence.py
```diff
@@ -143,6 +161,10 @@ def validate_positions(cls, v: Any) -> Any:
                 return list(v)
         return v
 
+    @validator("tile_plan", pre=True)
+    def validate_tileplan(cls, v: Any) -> Union[dict, NoTile]:
```
tlambert03, line 165: this doesn't look necessary.  you don't _have_ to have a validator for each field (only if you want to do something special).  Here, you're just return the default if `v` is Falsey... but that's exactly what the `default_factory` you specified above already does.   this can go

src/useq/_mda_sequence.py
```diff
@@ -166,6 +188,16 @@ def validate_mda(cls, values: Dict[str, Any]) -> Dict[str, Any]:
                 z_plan=values.get("z_plan"),
                 stage_positions=values.get("stage_positions", ()),
                 channels=values.get("channels", ()),
+                tile_plan=values.get("tile_plan"),
+            )
+
+        if not values.get("stage_positions", ()) and isinstance(
+            values.get("tile_plan"), TileRelative
+        ):
+            raise ValueError(
```
tlambert03, line 197: hmmm... we don't demand this for z-stack range around do we?  I think you can just assume the position is (0, 0) ... or check what we do in the case of range around without a position

src/useq/_mda_sequence.py
```diff
@@ -183,6 +215,7 @@ def _check_order(
         z_plan: Optional[AnyZPlan] = None,
         stage_positions: Sequence[Position] = (),
         channels: Sequence[Channel] = (),
+        tile_plan: Optional[AnyTilePlan] = None,
```
tlambert03, line 218: this doesn't appear to be used anywhere in this function?

src/useq/_mda_sequence.py
```diff
@@ -323,21 +357,41 @@ def iter_sequence(sequence: MDASequence) -> Iterator[MDAEvent]:
...
             continue
 
         _ev = dict(zip(order, item))
+
         index = {k: _ev[k][0] for k in INDICES if k in _ev}
 
         position: Optional[Position] = _ev[POSITION][1] if POSITION in _ev else None
         channel: Optional[Channel] = _ev[CHANNEL][1] if CHANNEL in _ev else None
         time: Optional[int] = _ev[TIME][1] if TIME in _ev else None
+        tile: Optional[dict[str, bool | float]] = _ev[TILE][1] if TILE in _ev else None
```
tlambert03, line 372: `dict[str, bool | float]` isn't all that helpful from a typing perspective... (i.e. giving the Union of all possible value types isn't very helpful since you can't do much with it ... either just say `dict` and leave it as is or use a TypedDict

src/useq/_mda_sequence.py
```diff
@@ -323,21 +357,41 @@ def iter_sequence(sequence: MDASequence) -> Iterator[MDAEvent]:
...
         if channel and TIME in index and index[TIME] % channel.acquire_every:
             continue
 
+        x_pos = getattr(position, "x", None)
+        y_pos = getattr(position, "y", None)
+
+        if tile:
+            if tile.get("is_relative"):
+                # e.g. TileRelative
+                dx = tile.get("dx")
```
tlambert03, line 386: don't use `.get` in a dict unless `None` is a valid value in later lines.  Here, it's not... since you can't add `x_pos` to `None`.  If the value _must_ be there use `tile['dx']`

tlambert03: also... I would just use `x` for both relative and not relative.  We already have the "is_relative" key that tells us the meaning of x ... we don't also need to rename it

src/useq/_mda_sequence.py
fdrgsp: for the z we assume it is 0 at the moment.

```python
if self.z_plan.is_relative:
    # TODO: either disallow without position z, or add concept of "current"
    z_pos += getattr(position, Z, None) or 0
```

But I thought it could be good to raise an error for both the tile and the z to avoid any stage movement that could potentially do some damage...even if for z=0 for inverted microscopes usually means that the objective is far for the sample - but what about right microscope? I'm not sure if 0 is far or close to the sample. And maybe for an XY stage it can be the same, where is (0, 0)?

src/useq/_mda_sequence.py
fdrgsp: That's true!

src/useq/_mda_sequence.py
tlambert03: > But I thought it could be good to raise an error for both the tile and the z to avoid any stage movement that could potentially do some damage..

you're thinking too much about the runtime behavior :) 

this library is an implementation agnostic *schema*.  It's not a python library for driving acquisition safely.  It knows nothing about how this experiment is run, and shouldn't be opinionated like that.  That kind of error should go in the thing that _runs_ an mda sequence ... or, you could also put that sort of check in the user interface that builds the MDA sequence

tlambert03: I'd say use the same "0" pattern as we do for z for now.  I can definitely imagine a downstream library making good use of that.  we shouldn't block it here

src/useq/_mda_sequence.py
fdrgsp: actually I think this is necessary in `validate_mda` that is validating the full MDASequence, am I wrong?

```python
@root_validator
def validate_mda(cls, values: Dict[str, Any]) -> Dict[str, Any]:
    if "axis_order" in values:
        values["axis_order"] = cls._check_order(
            values["axis_order"],
            z_plan=values.get("z_plan"),
            stage_positions=values.get("stage_positions", ()),
            channels=values.get("channels", ()),
            tile_plan=values.get("tile_plan"),
        )
    return values
```

src/useq/_mda_sequence.py
fdrgsp: at the moment we do that:
```python
x_pos = x_pos + dx if x_pos else None
y_pos = y_pos + dy if y_pos else None
```
Do you think it is ok or we need to make `x_pos` and `y_pos` = `0` if there is not a `stage_position`?

fdrgsp: actually it should be like that:
```python
x_pos = x_pos + dx if x_pos is not None else None
y_pos = y_pos + dy if y_pos is not None else None
```
but we can do like this instead, as you said it might be better:

```python
if tile:
    x_pos, y_pos = (x_pos or 0, y_pos or 0)
    if tile.get("is_relative"):
        # e.g. TileRelative
        dx = tile["x"]
        dy = tile["y"]
        x_pos = x_pos + dx
        y_pos = y_pos + dy
    else:
        # e.g. TileFromCorners
        x_pos = tile.get("x")
        y_pos = tile.get("y")
```

tests/conftest.py
```diff
@@ -28,4 +28,5 @@ def mda1() -> MDASequence:
             {"duration": {"minutes": 40}, "interval": 10},
         ],
         z_plan={"range": 1.0, "step": 0.5},
+        tile_plan={"is_relative": True, "rows": 2, "cols": 1},
```
tlambert03, line 31:
```suggestion
        tile_plan={"rows": 2, "cols": 1},
```

(`is_relative` is not a model field)

tlambert03 (approved):
great job here @fdrgsp!

PR #74, opened 2023-01-26
fix: fix TileRelative and TileFromCorners

fdrgsp:
in this PR:
- fix the  sign in `TileRelative` `_offset_y` method when `RelativeTo` is set to `center`.  If left negative, the grid is not calculated using the starting position as the center one.
- fix `TileFromCorners` so that the `corner` position are considered at the center of the image (and not at the top left corner).
- add specific tests for tiles.


tlambert03 (commented):
this should include a test

tlambert03 (approved):

PR #80, opened 2023-02-02
fix: update GridFromCorners with top, left, bottom, right floats

fdrgsp:
This PR updates the `GridFromCorners` object. Instead of setting up the grid bounding box using 2 coordinates (`corner1` and `corner2`), the grid is now defined by `top`, `left`, `bottom`, `right` `floats`.

### DIFF
--- a/src/useq/_grid.py
+++ b/src/useq/_grid.py
@@ -170,21 +170,21 @@ class GridFromEdges(_GridPlan):
     Attributes
     ----------
     top : float
-        y value of the top left bounding coordinate
+        value of the top bounding coordinate
     left : float
-        x value of the top left bounding coordinate
+        value of the left bounding coordinate
     bottom : float
-        y value of the bottom right bounding coordinate
+        value of the bottom bounding coordinate
     right : float
-        x value of the bottom right bounding coordinate
+        value of the right bounding coordinate
 
     Values are considered to be in the center of the image.
     """
 
-    top: float  # top_left y
-    left: float  # top_left x
-    bottom: float  # bottom_right y
-    right: float  # bottom_right x
+    top: float
+    left: float
+    bottom: float
+    right: float
 
     def _nrows(self, dy: float) -> int:
         total_height = abs(self.top - self.bottom) + dy
--- a/src/useq/_grid.py
+++ b/src/useq/_grid.py
@@ -135,11 +135,11 @@ class _GridPlan(FrozenModel):
     def _offset_y(self, dy: float) -> float:
         raise NotImplementedError
 
-    def _nrows(self, dx: float) -> int:
+    def _nrows(self, dy: float) -> int:
         """Return the number of rows, given a grid step size."""
         raise NotImplementedError
 
-    def _ncolumns(self, dy: float) -> int:
+    def _ncolumns(self, dx: float) -> int:
         """Return the number of columns, given a grid step size."""
         raise NotImplementedError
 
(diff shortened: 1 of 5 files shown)


src/useq/_grid.py
```diff
@@ -189,35 +164,41 @@ def _step_size(self, fov_width: float, fov_height: float) -> Tuple[float, float]
...
-        Second bounding coordinate (e.g. "bottom right"). The position is considered
-        to be in the center of the image.
+    top : float
+        y value of the top left bounding coordinate
+    left : float
+        x value of the top left bounding coordinate
+    bottom : float
+        y value of the bottom right bounding coordinate
+    right : float
+        x value of the bottom right bounding coordinate
```
tlambert03, line 179: fix all of these... the top doesn't care whether it's left/right.  the lefty doesn't care whether it's top/bottom, etc... it's just "top bound", "bottom bound", etc...

src/useq/_grid.py
```diff
@@ -189,35 +164,41 @@ def _step_size(self, fov_width: float, fov_height: float) -> Tuple[float, float]
...
+
+    Values are considered to be in the center of the image.
     """
 
-    corner1: Coordinate
-    corner2: Coordinate
+    top: float  # top_left y
```
tlambert03, line 184: remove comments... it's not top left, it's just top

src/useq/_grid.py
```diff
@@ -189,35 +164,41 @@ def _step_size(self, fov_width: float, fov_height: float) -> Tuple[float, float]
...
+    top : float
+        y value of the top left bounding coordinate
+    left : float
+        x value of the top left bounding coordinate
+    bottom : float
+        y value of the bottom right bounding coordinate
+    right : float
+        x value of the bottom right bounding coordinate
+
+    Values are considered to be in the center of the image.
```
tlambert03, line 181: not sure what this means... i would assume the values were at the the edges of the image

fdrgsp: > not sure what this means... i would assume the values were at the the edges of the image

it means that the value is considered to be a position at the center of the FOV and not for example the top left...Then in the general bounding box the top value will indeed define the top edge of the box. Does it make any sense?

tlambert03: ahhh, I think I see what you mean.  you're saying that when someone has a live view on, and they "mark the top" it's whatever stage position that's at... not the stage position - 1/2 FOV?

If so, I think that needs a fuller explanation, perhaps an example, etc...

tlambert03: and actually... and argument could be made that useq schema should _not_ care about that. And that that math should be done before passing to useq (or that we have a special constructor for that or something).  a bounding box is a simple, clear concept... it's a nice model.  a bounding box taking into consideration some as-of-yet unknown field of view is not so much a good model.

tlambert03: (sorry, one more thought):  just say that the position is the stage position (and not relative to an image) rather than saying it's at the center of the FOV.

fdrgsp: what about something like this?

```python
class GridFromEdges(_GridPlan):
    """Yield absolute stage positions to cover a bounded area...
    
    ...defined by setting its top, left, bottom and right
    edges stage coordinates.

    Attributes
    ----------
    top : float
        stage position of the top bounding area
    left : float
        stage position of the left bounding area
    bottom : float
        stage position of the bottom bounding area
    right : float
        stage position of the right bounding area
    """
```

tlambert03: Sure :) but make them "... of the bounding area".  Currently reads like there are four areas

tlambert03: Or just make them "top bounds", "left bounds", etc... and then say "all postitions in stage coordinates"

PR #81, opened 2023-02-03
feat: add sequence to Position object + update MDASequence iter_sequence + test [WIP]

fdrgsp:
This PR introduces a possible way to add an optional `MDASequence` to the `Position` object (instead of only the `z_plan` option) so that specific axes events (t, z, c or g) can be executed only for a specific position. It also updates the `MDASequence` `iter_sequence` method to take this `Position` sequence into account.

marge after #80


ianhi:
I'm a little bit confused. Can you post a few examples constructing these in the use cases you intend?

src/useq/_mda_sequence.py
```diff
@@ -344,51 +418,55 @@ def iter_sequence(sequence: MDASequence) -> Iterator[MDAEvent]:
...
+
+            p_event_iterator = (
+                enumerate(p_sequence.iter_axis(ax)) for ax in p_sequence.used_axes
+            )
+            for p_item in product(*p_event_iterator):
+                yield _iter_position_sequence(
+                    sequence, p_sequence, index, p_item, position, global_index
+                )
+                global_index += 1
+            continue
```
tlambert03, line 458: notice how `_iter_position_sequence` is pretty much the same as `iter_sequence`?  use recursion! :)

```suggestion
        if position and position.sequence:
            for sub_event in iter_sequence(position.sequence):
                subz = (z_pos or 0) + sub_event.z_pos if sub_event.z_pos else z_pos
                subx = (x_pos or 0) + sub_event.x_pos if sub_event.x_pos else x_pos
                suby = (y_pos or 0) + sub_event.y_pos if sub_event.y_pos else y_pos
                yield sub_event.copy(
                    update={
                        "global_index": global_index,
                        "index": {**index, **sub_event.index},
                        "x_pos": subx,
                        "y_pos": suby,
                        "z_pos": subz,
                        "sequence": sequence,
                    }
                )
                global_index += 1
            continue
```

and then delete the `_iter_position_sequence` function below.

then actually, that whole bit about `subz`, `subx` etc... could be a method on the MDAEvent class to essentially "shift" its position a bit, returning a new MDA event object

src/useq/_mda_sequence.py
```diff
@@ -401,3 +479,42 @@ def iter_sequence(sequence: MDASequence) -> Iterator[MDAEvent]:
             sequence=sequence,
             global_index=global_index,
         )
+        global_index += 1
+
+
+def _iter_position_sequence(
```
tlambert03, line 485: not actually an iterator (just returns a single value) ... so would want a different name, but it will be removed anyway, so no biggie

src/useq/_mda_sequence.py
```diff
@@ -119,6 +126,12 @@ def set_fov_size(self, fov_size: Tuple[float, float]) -> None:
         This is used to calculate the number of positions in a grid plan.
         """
         self._fov_size = fov_size
+        # set the same fov also to any position sequence
+        for pos in self.stage_positions:
```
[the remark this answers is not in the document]
tlambert03: tests required to demonstrate what it is you're trying to fix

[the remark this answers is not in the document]
fdrgsp: removed, is not used anymore

src/useq/_mda_sequence.py
```diff
@@ -32,7 +42,6 @@ class MDASequence(UseqModel):
     This is the core object in the `useq` library, and is used define a sequence of
     events to be run on a microscope. It object may be constructed manually, or from
     file (e.g. json or yaml).
-
```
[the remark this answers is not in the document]
tlambert03: these comments are all resolved, but the lines are still touched.  Add back the newlines

src/useq/_mda_sequence.py
```diff
@@ -73,27 +82,25 @@ class MDASequence(UseqModel):
     ... )
     >>> print(seq)
     Multi-Dimensional Acquisition ▶ nt: 2, np: 1, nc: 1, nz: 4, ng: 4
-
```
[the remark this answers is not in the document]
tlambert03: revert

src/useq/_mda_sequence.py
```diff
@@ -73,27 +82,25 @@ class MDASequence(UseqModel):
...
     >>> for event in seq:
     ...     print(event)
-
```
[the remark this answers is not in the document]
tlambert03: revert

src/useq/_mda_sequence.py
```diff
@@ -73,27 +82,25 @@ class MDASequence(UseqModel):
...
-    grid_plan:
-      cols: 2
-      rows: 2
     time_plan:
       interval: '0:00:00.100000'
       loops: 2
     z_plan:
       range: 3.0
-      step: 1.0
+      step: 1.0.
```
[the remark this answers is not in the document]
tlambert03: this is still there.  fix it

tests/fixtures/mda.yaml
```diff
@@ -13,16 +13,35 @@ grid_plan:
...
   some info: something
 stage_positions:
-- x: 10.0
+- name: null
+  sequence: null
+  x: 10.0
   y: 20.0
+  z: null
 - name: test_name
+  sequence:
```
tlambert03, line 22: for the `mda.yaml` fixture, leave out all the fields that are just their default values (like `channels: []` or `go_up: true`, etc...).  they have to be in the json fixture, but not this one

src/useq/_mda_sequence.py
```diff
@@ -375,20 +390,27 @@ def iter_sequence(sequence: MDASequence) -> Iterator[MDAEvent]:
...
-        except sequence._SkipFrame:
+        if position and position.sequence:
+            for sub_event in iter_sequence(position.sequence):
+                if (
+                    position.sequence.grid_plan
+                    and position.sequence.grid_plan.is_relative
+                ):
+                    sub_event = sub_event.shifted(x_pos=x_pos, y_pos=y_pos)
+
+                if position.sequence.z_plan and position.sequence.z_plan.is_relative:
```
tlambert03, line 401: what about subpositions with absolute plans?  imagine each position specified a different Z-top-bottom.  i think that would have worked before but not now?

add a test with two subpositions having `z_plan: {top: ..., bottom: ...}`

tlambert03 (changes requested):

tests/test_position_sequence.py
```diff
@@ -0,0 +1,147 @@
...
+@pytest.fixture
+def mda() -> MDASequence:
+    return MDASequence(
+        axis_order="tpgcz",
+        stage_positions=[(10, 20, 0), {"name": "test", "x": 10, "y": 20, "z": 50}],
+        channels=[{"config": "Cy5", "exposure": 50}],
+    )
+
+
+def _update_pos_sequence(mda: MDASequence, axis: str):
```
tlambert03, line 15: i'm not a fan of this pattern... it requires someone reading the tests to jump all over the place.  You only ever use each axis argument ("c", "z", etc...) once in the tests here, so do the replacement in the test.  then it's more obvious to the reader what the input of the test actually was

tests/test_position_sequence.py
```diff
@@ -0,0 +1,147 @@
...
+    if "t" in axis:
+        p_seq = p_seq.replace(time_plan=[{"interval": 1, "loops": 2}])
+    new_pos = [
+        (10, 20, 0),
+        {"name": "test", "x": 10, "y": 20, "z": 50, "sequence": p_seq},
+    ]
+    return mda.replace(stage_positions=new_pos)
+
+
+def event_list(sequence: MDASequence) -> list:
```
tlambert03, line 32: i also think this makes the tests hard to read... you have to remember what order you picked up here (global index, then index, then pos_name) etc... Making the tests themselves very hard to look at

tests/test_position_sequence.py
tlambert03: also, you don't *have* to use a fixture for the original mda.  Rather than creating a very small `mda` fixture, that you then selective replace in each test.  Just create the mda you want to test inside of each test

tests/test_position_sequence.py
```diff
@@ -0,0 +1,147 @@
...
+            i.min_start_time,
+            i.exposure,
+            # i.channel,
+        )
+        for i in sequence.iter_events()
+    ]
+
+
+def test_position_sequence_channels(mda: MDASequence) -> None:
+    mda1 = mda.replace(
```
tlambert03, line 50: see above.  if you're just immediately going to override the fixture, don't use it.  make a new one here

tlambert03 (commented):
i think I'm still looking for something a bit more obvious explicit in the tests here:  For example, it might reflect the conversation we were having.

i.e. 
```python

# test that z positions are overridden (position 1), 
# relatively shifted (position 2),
# or inherited from the parent sequence (position 3). 
MDASequence(
    z_plan={"top": 60, "bottom": 55, "step": 1},
    stage_positions=[
        {"x": 40, "y": 50, "z": 60, "sequence": {"z_plan": {"top": 20, "bottom": 0}}},
        {"x": 60, "y": 70, "z": 80, "sequence": {"z_plan": {"range": 20, "step": 0.5}}},
        {"x": 10, "y": 20, "z": 30},
    ],
)
```

... and you can do a similar thing for channel, grid, etc...

tlambert03 (approved):
awesome work here @fdrgsp!

PR #91, opened 2023-06-30
feat: add Autofocus Plan

fdrgsp:
This PR adds an **_hardware autofocus device_** plan to `useq-schema` and implements its use in the `MDASequence` `iter_sequence` method (and `MDAEvent`). 

The basic logic is that a user can specify the `axes` that should be used for the hardware autofocus (e.g. `p`, `t`, etc.) and the yielded `MDAEvent` will have info on weather to perform autofocus or not on the specified axis.

merge after #93 .

Example of use in https://github.com/pymmcore-plus/pymmcore-plus/pull/199 and https://github.com/pymmcore-plus/pymmcore-widgets/pull/147.


src/useq/_mda_event.py
```diff
@@ -35,10 +35,16 @@ class Channel(UseqModel):
...
         `"Channel"`.
+    do_stack : bool
+        If `True`, include this channel in any z stacks acquired. By default, `True`.
+    z_offset : float
+        Relative Z offset from z position, in microns. By default, `0`.
     """
 
     config: str
     group: str = "Channel"
+    do_stack: bool = True
```
tlambert03, line 46: remove

src/useq/_mda_event.py
```diff
@@ -98,6 +104,8 @@ class MDAEvent(UseqModel):
     z_pos : float | None
         Z position in microns. If not provided, implies use current position. By
         default, `None`.
+    autofocus : tuple[str, float] | None
```
tlambert03, line 107: "When not None, consumers of this event must perform a hardware autofocus as follows: ..."

src/useq/_mda_event.py
```diff
@@ -127,6 +135,7 @@ class MDAEvent(UseqModel):
     sequence: Optional[MDASequence] = Field(default=None, repr=False)
     global_index: int = Field(default=0, repr=False)
     metadata: Dict[str, Any] = Field(default_factory=dict)
+    autofocus: Optional[Tuple[str, float]] = None
```
tlambert03, line 138: make a new object, like Channel has

tlambert03:
as discussed, perhaps a top level `AutoFocusPlan` on the sequence itself is a good idea here:

```py
class AutoFocusPlan(UseqModel):
    device_name: str
    af_motor_position: float = 0
    axes: Tuple[str, ...] = Field(default_factory=tuple)

```

src/useq/_mda_event.py
tlambert03:
```
class AutoFocus(UseqModel):
    device_name: str
    af_motor_position: float
    focus_motor_position: float
```

PR #96, opened 2023-07-15
feat: add Actions to MDA event, and add autofocus events

tlambert03:
here ya go @fdrgsp  :) 
ended up being only a few line change in iter_sequence.  have a look and we can talk about it soon.
I updated the test patterns too.  I didn't fully reimplement all of your scenarios, but I'm confident they'll work, since at this point, all we're really testing is just the logic of the very simple  `def should_autofocus(self, event: MDAEvent) -> bool:` method.

closes #91, closes #94

this includes most of #94, and a little cleanup is still needed


src/useq/_mda_sequence.py
```diff
@@ -450,24 +468,38 @@ def iter_sequence(
...
+            # and we don't hit the continue statement, but we can use the autofocus plan
+            elif position.sequence is not None and position.sequence.autofocus_plan:
+                autofocus_plan = position.sequence.autofocus_plan
+
+        event = MDAEvent(**event_kwargs)
+        if autofocus_plan:
+            af_event = autofocus_plan.event(event)
+            if af_event is not None:
+                yield af_event
+        yield event
```
tlambert03, line 502: this is the bulk of it... the rest is just an additional indentation

fdrgsp: @tlambert03 that's fantastic :tada: !!!

I guess the only thing left in terms of the logic is to update the `z_pos` of the autofocus event in case of a relative `z_plan` (this way the z stage will move at the home position of the z stack) like we were doing here for example https://github.com/fdrgsp/useq-schema/blob/9584af6a21dde61da74a9405db467e47ab5c7f02/src/useq/_mda_sequence.py#L549-L550.

for example
```py
event = MDAEvent(**event_kwargs)
if autofocus_plan:
    af_event = autofocus_plan.event(event)
    if af_event is not None:
        yield af_event.copy(
            update={
                "z_pos": _get_updated_z_pos(
                    event.z_pos,
                    event.index,
                    position.sequence if position else None,
                    sequence,
                )
            }
        )
```

and add 3 simple tests like these:

```py
def test_autofocus_z_pos() -> None:
    mda = TWO_CH.replace(stage_positions=[ZPOS_200], z_plan=ZRANGE2)
    af_plan = AxesBasedAF(
        autofocus_device_name="Z", autofocus_motor_offset=50, axes=("c",)
    )
    mda = mda.replace(autofocus_plan=af_plan)
    z = [e.z_pos for e in mda]
    assert z[0] == z[4] == 200


def test_autofocus_z_pos_af_sub_sequence() -> None:
    mda = TWO_CH.replace(stage_positions=[SUB_P_AF_C], z_plan=ZRANGE2)
    z = [e.z_pos for e in mda]
    assert z[0] == z[4] == 10


def test_autofocus_z_pos_af_sub_sequence_zplan() -> None:
    mda = TWO_CH_SUBPAF_C
    z = [e.z_pos for e in mda]
    assert z[2] == z[4] == 10
```

tlambert03: > `_get_updated_z_pos`

sounds good.  do you want to take a stab at that and PR against this PR? (i.e. my `merge-autofocus` branch?)

- I think the logic should go in the `AutoFocusPlan.event` method, and it should only be added to the `copy(update=` dict if necessary.
- I don't think we'll need the `sequence`/`main_sequence` anymore... it should hopefully just be at `event.sequence` by that point

tests/test_autofocus.py
```diff
@@ -0,0 +1,113 @@
...
+        stage_positions=[ZPOS_200],
+        z_plan=useq.ZAboveBelow(above=2, below=2, step=1),
+        autofocus_plan=AF_C,
+    )
+    assert all(e.z_pos == 200 for e in mda if isinstance(e.action, HardwareAutofocus))
+
+
+def test_autofocus_z_pos_af_sub_sequence() -> None:
+    mda = TWO_CH.replace(stage_positions=[SUB_P_AF_C], z_plan=ZRANGE2)
+    assert all(e.z_pos == 10 for e in mda if isinstance(e.action, HardwareAutofocus))
```
tlambert03, line 97: quick note @fdrgsp  ... note how this is much more interpretable to a reader:

```python
assert all(e.z_pos == 10 for e in mda if isinstance(e.action, HardwareAutofocus))
```
than this:

```python
z = [e.z_pos for e in mda]
assert z[2] == z[4] == 10
```

PR #97, opened 2023-07-17
feat: add keep_shutter_open logic

fdrgsp:
@tlambert03 for the `MDAEvent` `keep_shutter_open` what do you think about using a logic similar to the autoshutter?

This PR gives an example of that.

Briefly, the MDASequence has a `shutter_plan` which is a `ShutterOpenAxes` class that can be used to specify the axes where the shutter should be kept open.
As the `AutoFocusPlan`, `ShutterOpenAxes` has a `should_keep_open` method that return True if the specified axes have changed from the previous event.

So we can simply call `shutter_plan.should_keep_open(event)` before yielding the event to set the `MDAEvent` `keep_shutter_open` attribute to True or False.
```py
if shutter_plan.should_keep_open(event):
    event = event.copy(update={"keep_shutter_open": True})
```


src/useq/_mda_event.py
```diff
@@ -115,6 +115,8 @@ class MDAEvent(UseqModel):
         The action to perform for this event.  By default, [`useq.AcquireImage`][].
         Example of another action is [`useq.HardwareAutofocus`][] which could be used
         to perform a hardware autofocus.
+    keep_shutter_open : bool
+        Keep the shutter open between during this event. By default, `False`.
```
tlambert03, line 119: what does "between during this event" mean?  Does it mean "Do not close the illumination shutter at the end of this event?"

src/useq/_mda_sequence.py
```diff
@@ -420,12 +427,16 @@ def iter_sequence(
         A dictionary of offsets to apply to each position. This can be used to shift
         all positions in a sub-sequence.  Keys must be one of `x_pos`, `y_pos`, or
         `z_pos` and values should be floats.s
+    shutter_plan : ShutterOpenAxes | None
```
tlambert03, line 430: remove

src/useq/_mda_sequence.py
```diff
@@ -484,15 +495,23 @@ def iter_sequence(
                     base_event_kwargs=event_kwargs.copy(),
                     event_kwarg_overrides=pos_overrides,
                     position_offsets=_offsets,
+                    shutter_plan=(
+                        position.sequence.shutter_plan or sequence.shutter_plan
+                    ),
```
tlambert03, line 500: remove, use the same approach that we did for autofocus_plan

src/useq/_mda_event.py
fdrgsp: it's a mistake, it should be updated as:

`If True, the shutter should be kept open during this event. By default, False.`

src/useq/_mda_sequence.py
```diff
@@ -484,15 +495,23 @@ def iter_sequence(
...
                 )
                 continue
             # note that position.sequence may be Falsey even if not None, for example
             # if all it has is an autofocus plan.  In that case, we don't recurse.
             # and we don't hit the continue statement, but we can use the autofocus plan
             elif position.sequence is not None and position.sequence.autofocus_plan:
                 autofocus_plan = position.sequence.autofocus_plan
+                shutter_plan = position.sequence.shutter_plan or sequence.shutter_plan
```
tlambert03, line 508: _this_ should be the only thing you need. no?  we didn't need all the other stuff for autofocus plan, so I'm confused why we would need it for shutter plan

tlambert03, line 508: _this_ should be the only thing you need. no?  we didn't need all the other stuff for autofocus plan, so I'm confused why we would need it for shutter plan

src/useq/_shutter_plan.py
```diff
@@ -0,0 +1,37 @@
...
+        is change, (in other words: every time the channel is changed.).
+    """
+
+    axes: Tuple[str, ...] = ()
+    _previous: dict = PrivateAttr(default_factory=dict)
+
+    def __bool__(self) -> bool:
+        return bool(self.axes)
+
+    def should_keep_open(self, event: MDAEvent) -> bool:
```
tlambert03, line 27: since this is the same thing as the other function, we should reuse it

tests/test_keep_shutter_open.py
```diff
@@ -0,0 +1,89 @@
+from __future__ import annotations
+
+from typing import Iterable
+
+import pytest
+
+import useq
+from useq import AxesBasedAF, MDASequence, ShutterOpenAxes
+
+ZRANGE2 = useq.ZRangeAround(range=2, step=1)
```
tlambert03, line 10: why do I get the feeling that you just copied and pasted everything from the autofocus one?  :)

this is wayyy to much.  all you're really testing here is that `should_keep_open` works as it should. and all it's testing is that if a "watched" axis changes you return True.  we probably didn't need all these cases for autofocus in the first place after all the logic was cleaned up, and we shouldn't continue to propogate it.

src/useq/_mda_sequence.py
tlambert03, line 508: _this_ should be the only thing you need. no?  we didn't need all the other stuff for autofocus plan, so I'm confused why we would need it for shutter plan

src/useq/_mda_sequence.py
tlambert03, line 430: remove

src/useq/_mda_sequence.py
```diff
@@ -420,12 +427,16 @@ def iter_sequence(
...
+    shutter_plan : ShutterOpenAxes | None
+        The plan for keeping the shutter open for any specified axes. By default, None.
+        It will default to the `shutter_plan` of the sequence, if present.
 
     Yields
     ------
     MDAEvent
         Each event in the MDA sequence.
     """
+    shutter_plan = shutter_plan or sequence.shutter_plan
```
tlambert03, line 439: remove

src/useq/_mda_sequence.py
```diff
@@ -388,6 +394,7 @@ def iter_sequence(
     base_event_kwargs: MDAEventDict | None = None,
     event_kwarg_overrides: MDAEventDict | None = None,
     position_offsets: PositionDict | None = None,
+    shutter_plan: ShutterOpenAxes | None = None,
```
tlambert03, line 397: already add stuff to our nice and trimmed down signature? ;)

this isn't necessary.  see note below

src/useq/_mda_sequence.py
tlambert03, line 439: remove

src/useq/_mda_sequence.py
tlambert03, line 397: already add stuff to our nice and trimmed down signature? ;)

this isn't necessary.  see note below

docs/schema/shutter_plan.md
```diff
@@ -0,0 +1,15 @@
...
+
+Keep the shutter open along the specified axes.
+
+::: useq.ShutterOpenAxes
+    options:
+        show_signature: true
+        show_signature_annotations: true
+        members: []
+
+::: useq.ShutterOpenAxes.should_keep_open
```
tlambert03, line 11: i don't think we need to document the python implementation stuff here.  Just the schema

tlambert03:
that confuses me too... What does "during this event" mean?  Aren't shutters almost always open during an event?  Remember, an event tends to be a single snap.  I think what you're trying to express is that the shutter should remain open even when this event is done.  For example, if someone wanted to keep it open throughout a z stack (most common thing to do) then they would want the shutter to remain open during *many events* (not just one).  I think you know that, but the wording is still unclear.

tlambert03:
i think i'm ok with this.  I'm a little hesitant to keep adding "plans" to the top level of the MDASequence schema.  Since this is rather similar to autofocus (with the primary difference being it doesn't *add* events, it *modifies* events...) I'd like to think for a moment about what the general patterns are and think about whether this can be done in a slightly more general way

fdrgsp:
> that confuses me too... What does "during this event" mean? Aren't shutters almost always open during an event? Remember, an event tends to be a single snap. I think what you're trying to express is that the shutter should remain open even when this event is done. For example, if someone wanted to keep it open throughout a z stack (most common thing to do) then they would want the shutter to remain open during *many events* (not just one). I think you know that, but the wording is still unclear.
> […](#)


Yes I agree. I thought that every event should have a `keep_shutter_open` bool so that an engine will check if it has to set the shutter to the open state of the closed state.

for example, in case of our `pymmcore_plus` engine we could use the `FullPMDAEngine` and have something like:

```py
def setup_event(self, event: MDAEvent) -> None:
        ...
        if self._use_autoshutter and event.keep_shutter_open:
            self._mmc.setAutoShutter(False)
            self._mmc.setShutterOpen(True)

        self._mmc.waitForSystem()
```

and then, at the end of a sequence we can reset the shutter to its original state (by original I mean the state when the sequence started).
```py 
def teardown_sequence(self, sequence: MDASequence) -> None:
        """Teardown state of system (hardware, etc.) after `sequence`."""
        if self._use_autoshutter:
            self._mmc.setAutoShutter(True)
            self._mmc.setShutterOpen(False)

**Comment by fdrgsp** [comment]
> i think i'm ok with this. I'm a little hesitant to keep adding "plans" to the top level of the MDASequence schema. Since this is rather similar to autofocus (with the primary difference being it doesn't _add_ events, it _modifies_ events...) I'd like to think for a moment about what the general patterns are and think about whether this can be done in a slightly more general way

Sure! I had the same feeling but I thought it could a starting point.

File: src/useq/_mda_sequence.py (hunk 8 above)
Comment (fdrgsp) [reply src/useq/_mda_sequence.py:397]: Actually I found an autofocus bug in the code and I think we should add a `autofocus_plan_override` argument to the `iter_sequence` method. I opened this PR  #98  to explain it in detail.

Comment (tlambert03) [reply src/useq/_mda_sequence.py:397]: I still don't think so but I'll have a look.  We have kwargs for both fallbacks and overrides ... so it's hard to see how we
Need more. (It's possible you need to change the arguments passed in to either the fallbacks or overrides, but I still don't think we need more kwargs)

Comment (tlambert03) [reply src/useq/_mda_sequence.py:397]: Yeah I think what you want to do is use sequence.replace on the sequence being passed.  actually, I think we could do that for all the overrides and possibly the fallbacks too come to think of it.  Getting us back to just a single argument

**Comment by tlambert03** [comment]
> Yes I agree. I thought that every event should have a `keep_shutter_open` bool so that an engine will check if it has to set the shutter to the open state of the closed state.


yep. I agree with all of this, my point is just that the docs are confusing.  It's not what happens during the event, it's what happens after it.   The docs here should make it super clear to someone implementing the engine exactly what they are supposed to do.  I think we're in agreement about what `keep_shutter_open` means, but it needs to be made clearer.  I would say "If `keep_shutter_open` is True, the illumination shutter should be left open after the event has been executed, otherwise it should be closed."

**Comment by tlambert03** [comment]
having another look at this... i'm not sure we want to reuse the same logic from autofocus here.  Or, maybe we do but I'm not getting it yet...  Can you give me an example of how you'd express "keep the shutter open while moving z", in a multi-channel, z stack?  Particularly because this is more of a thing that persists *across* events (or rather,  an instruction for how to leave the state), it feels different.

Also shouldn't we be ensuring that we're only keeping the shutter open if the "keep open" axis is the fastest moving one?  For example, you can't keep the illumination shutter open if the channel is changing.  So there appear to be more limitations here than autofocus.  In general, does it ever make sense to have the kept-open axis be slower moving than channel?

**Comment by tlambert03** [comment]
i think it would be helpful to delete the `test_keep_shutter_open` file and make some clearer tests, without just reusing the autofocus file.  (not sure for the sake of removing duplication, but since I'm not sure it's the same thing)

**Comment by fdrgsp** [comment]
>  Can you give me an example of how you'd express "keep the shutter open while moving z", in a multi-channel, z stack

So basically I was thinking that if you want to have the shutter open during a z stack, then you have to specify the "z" in the shutter_plan `shutter_plan={"axes":("z",)}`:

running this:
```py
s = MDASequence(
    channels=["DAPI", "FITC"],
    stage_positions=[(0,0,10)],
    shutter_plan={"axes":("z",)},
    z_plan={"range": 1.0, "step": 0.5}
)

for e in s:
    print(f"INDEX:' {e.index}, SHUTTER_OPEN: {e.keep_shutter_open}")
```
results in this:
```
INDEX: {'p': 0, 'c': 0, 'z': 0} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 0, 'z': 1} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 0, 'z': 2} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 1, 'z': 0} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 1, 'z': 1} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 1, 'z': 2} SHUTTER_OPEN: True
```
because the first axes changing is the "z" axes.

if you change the axis_order to `tpgzc`
```

s = MDASequence(
    axis_order="tpgzc",
    channels=["DAPI", "FITC"],
    stage_positions=[(0,0,10)],
    shutter_plan={"axes":("z",)},
    z_plan={"range": 1.0, "step": 0.5}
)
```
then you get
```
INDEX:' {'p': 0, 'c': 0, 'z': 0}, SHUTTER_OPEN: True
INDEX:' {'p': 0, 'c': 1, 'z': 0}, SHUTTER_OPEN: False
INDEX:' {'p': 0, 'c': 0, 'z': 1}, SHUTTER_OPEN: True
INDEX:' {'p': 0, 'c': 1, 'z': 1}, SHUTTER_OPEN: False
INDEX:' {'p': 0, 'c': 0, 'z': 2}, SHUTTER_OPEN: True
INDEX:' {'p': 0, 'c': 1, 'z': 2}, SHUTTER_OPEN: False
```
because the first axes changing is the channel one. So in this case, if you want to have the shutter always open you just have to set the  `shutter_plan` axes to both `("c", "z")`.


This is what I thought it could make sense :) but probably there is a different logic that works!

fdrgsp:
> i think it would be helpful to delete the `test_keep_shutter_open` file and make some clearer tests, without just reusing the autofocus file. (not sure for the sake of removing duplication, but since I'm not sure it's the same thing)

I used the same logic of the autofocus test first to have something fast to use as tests :) but also because since we have an extra autofocus event, I wanted to make sure that the shutter_plan works also during the autofocus events.

tlambert03:
yes, but practically speaking...


```python
INDEX: {'p': 0, 'c': 0, 'z': 0} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 0, 'z': 1} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 0, 'z': 2} SHUTTER_OPEN: True   # <- shouldn't this be False?
INDEX: {'p': 0, 'c': 1, 'z': 0} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 1, 'z': 1} SHUTTER_OPEN: True
INDEX: {'p': 0, 'c': 1, 'z': 2} SHUTTER_OPEN: True   # <- shouldn't this be False?
```

```python
# shouldn't these ALL be false?
INDEX:' {'p': 0, 'c': 0, 'z': 0}, SHUTTER_OPEN: True
INDEX:' {'p': 0, 'c': 1, 'z': 0}, SHUTTER_OPEN: False
INDEX:' {'p': 0, 'c': 0, 'z': 1}, SHUTTER_OPEN: True
INDEX:' {'p': 0, 'c': 1, 'z': 1}, SHUTTER_OPEN: False
INDEX:' {'p': 0, 'c': 0, 'z': 2}, SHUTTER_OPEN: True
INDEX:' {'p': 0, 'c': 1, 'z': 2}, SHUTTER_OPEN: False
```

tlambert03:
i think the point here is that, unlike autofocus events, leaving the shutter open  is intrinsically linked to the channel axis.  It doesn't ever make sense to leave the shutter open if the next image can't be acquired with that same "open shutter".  So, you're predicate here really needs awareness of whether the *next* event is using the same channel config as the current one

PR #98, opened 2023-07-17
fix: fix inherit sequence autofocus + add tests

fdrgsp:
I found an `autofocus_plan` bug in the `MDASequence` `iter_sequence` method and this PR is a possible fix (also related to #97 ).

If we have a `MDASequence` with a position sub-sequence (like the one below)  where the `autofocus_plan` is *only* specified in the main sequence, the `autofocus_plan` should be also inherited by the position sub-sequence. However at the moment this does not happen.

```py
mda = MDASequence(
    channels=["DAPI", "FITC"],
    stage_positions=[(1, 2, 3),{"z": 10, "sequence":{"z_plan": {"range": 2, "step": 1}}}],
    autofocus_plan={"autofocus_device_name": "Z", "autofocus_motor_offset": 40, "axes": ("z",)}
)
```

by printing index and action type
```
for e in mda:
    print(e.index, e.action.type
```
we get
```
{'p': 0, 'c': 0} acquire_image
{'p': 0, 'c': 1} acquire_image
{'p': 1, 'c': 0, 'z': 0} acquire_image
{'p': 1, 'c': 0, 'z': 1} acquire_image
{'p': 1, 'c': 0, 'z': 2} acquire_image
{'p': 1, 'c': 1, 'z': 0} acquire_image
{'p': 1, 'c': 1, 'z': 1} acquire_image
{'p': 1, 'c': 1, 'z': 2} acquire_image
```
indicating that the auto shutter is never used. While we should instead autofocus performed at every z planes like so
```
{'p': 0, 'c': 0} acquire_image
{'p': 0, 'c': 1} acquire_image
{'p': 1, 'c': 0, 'z': 0} hardware_autofocus
{'p': 1, 'c': 0, 'z': 0} acquire_image
{'p': 1, 'c': 0, 'z': 1} hardware_autofocus
{'p': 1, 'c': 0, 'z': 1} acquire_image
{'p': 1, 'c': 0, 'z': 2} hardware_autofocus
{'p': 1, 'c': 0, 'z': 2} acquire_image
{'p': 1, 'c': 1, 'z': 0} hardware_autofocus
{'p': 1, 'c': 1, 'z': 0} acquire_image
{'p': 1, 'c': 1, 'z': 1} hardware_autofocus
{'p': 1, 'c': 1, 'z': 1} acquire_image
{'p': 1, 'c': 1, 'z': 2} hardware_autofocus
{'p': 1, 'c': 1, 'z': 2} acquire_image
```

### DIFF
--- a/src/useq/_mda_sequence.py
+++ b/src/useq/_mda_sequence.py
@@ -382,13 +382,20 @@ class PositionDict(TypedDict, total=False):
     z_pos: float
 
 
+class MDASequenceDict(TypedDict, total=False):
+    autofocus_plan: AnyAutofocusPlan | None
+    # this could also include the shutter_plan to set MDAEvents keep_shutter_open
+    # argument, but that is not yet implemented
+    # see https://github.com/pymmcore-plus/useq-schema/pull/97
+
+
 def iter_sequence(
     sequence: MDASequence,
     *,
     base_event_kwargs: MDAEventDict | None = None,
     event_kwarg_overrides: MDAEventDict | None = None,
     position_offsets: PositionDict | None = None,
-    autofocus_plan_override: AnyAutofocusPlan | None = None,
+    sequence_kwargs_overrides: MDASequenceDict | None = None,
 ) -> Iterator[MDAEvent]:
     """Iterate over all events in the MDA sequence.'.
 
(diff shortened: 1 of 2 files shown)


tlambert03:
can you try to do this with 

```python
                sub_seq = position.sequence
                # if the sub-sequence doe not have an autofocus plan, we override it
                # with the parent sequence's autofocus plan
                if not sub_seq.autofocus_plan:
                    sub_seq = sub_seq.replace(autofocus_plan=sequence.autofocus_plan)

                # recurse into the sub-sequence
                yield from iter_sequence(
                    sub_seq,
                    ...
```

shouldn't that do the same thing?

tlambert03:
> can you try to do this with
> 
> ```python
>                 sub_seq = position.sequence
>                 # if the sub-sequence doe not have an autofocus plan, we override it
>                 # with the parent sequence's autofocus plan
>                 if not sub_seq.autofocus_plan:
>                     sub_seq = sub_seq.replace(autofocus_plan=sequence.autofocus_plan)
> 
>                 # recurse into the sub-sequence
>                 yield from iter_sequence(
>                     sub_seq,
>                     ...
> ```
> 
> shouldn't that do the same thing?

yeah works... but you need to use `copy()` not `replace` ... so as to maintain the internal state of the should_autofocus function.  see https://github.com/fdrgsp/useq-schema/pull/4

fdrgsp:
> yeah works... but you need to use `copy()` not `replace` ... so as to maintain the internal state of the should_autofocus function. see [fdrgsp#4](https://github.com/fdrgsp/useq-schema/pull/4)

Yesssss!Works well👍🏻

PR #99, opened 2023-07-18
fix: KeyError fix in MDASequence _should_skip method

fdrgsp:
this PR fixes a bug  in the `MDASequence` `_should_skip` method.

if we run this
```py
mda = MDASequence(channels=[{"config": "DAPI", "do_stack": False}])
    assert len(list(mda)) == 1
```
we get a "z" KeyError because in the `_should_skip` method we don't make sure to also check for a `z_plan`
```py
# only acquire on the middle plane:
if not channel.do_stack and index[Z] != len(z_plan) // 2:
    return True
```
note: just switching from `index[Z]` to `index.get(Z)` does not work because the event will be skipped even when it shouldn't be.

### DIFF
--- a/src/useq/_mda_sequence.py
+++ b/src/useq/_mda_sequence.py
@@ -565,7 +565,7 @@ def _should_skip(
             return True
 
         # only acquire on the middle plane:
-        if not channel.do_stack and index[Z] != len(z_plan) // 2:
+        if not channel.do_stack and z_plan and index[Z] != len(z_plan) // 2:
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
thanks!  

as a side thought, I think we should think about a generic `ZPlan.home_position()` method.  Rather than len(z_plan)//2 which wouldn't work for asymmetric ranges.  Would also be relevant for autofocus z

PR #105, opened 2023-07-21
fix: remove NoX plans from docs

fdrgsp:
Remove NoX


tlambert03:
thanks!  As a side-task, would you want to set-up CI at useq-schema to run widget tests, similar to [how we run pymmcore-plus tests](https://github.com/pymmcore-plus/useq-schema/blob/db42465d2c86405a9b20861af73c305de6bc244e/.github/workflows/test_and_deploy.yml#L52)?

PR #116, opened 2023-07-29
feat: Add FOV size to grid spec

tlambert03:
here it is @fdrgsp, the one you've always wanted :)


fdrgsp (approved):
I like this change :)!

PR #119, opened 2023-07-30
feat: add `keep_shutter_open_across` to MDASequence

### LINKED ISSUE #3: keep shutter open
in pycromanager, there is an event key called `keep_shutter_open` that instructs the hardware to keep the shutter(s?) open across C/Z.  should add that concept somewhere to the schema

see discussion here https://github.com/micro-manager/pycro-manager/issues/88

tlambert03:
This is alternative to #97, that adds a `keep_shutter_open` logic to the MDA sequence.  The logic is described in the docstring:

> A tuple of axes `str` across which the illumination shutter should be kept open.
        Resulting events will have `keep_shutter_open` set to `True` if and only if
        ALL axes whose indices are changing are in this tuple. For example, if
        `keep_shutter_open_across=('z',)`, then the shutter would be kept open between
        events axes {'t': 0, 'z: 0} and {'t': 0, 'z': 1}, but not between
        {'t': 0, 'z': 0} and {'t': 1, 'z': 0}.

closes #97 
closes #3

sidenote: this also opens up a possibility for a generalization of both this and autofocus plan, which would be general functions that take in the current event and (optionally) the upcoming event, and return either a modified current event, a new event or sequence of events to insert, or block yielding of the current event

### DIFF
--- a/tests/test_sequence.py
+++ b/tests/test_sequence.py
@@ -385,20 +385,14 @@ def test_keep_shutter_open() -> None:
         is_last_zt = bool(event.index["t"] == 1 and event.index["z"] == 1)
         assert event.keep_shutter_open != is_last_zt
 
-    # testing keep shutter open in sub-sequence
+    # even though c is the last axis, and comes after g, because the grid happens
+    # on a subsequence shutter will be open across the grid for each position
+    subseq = MDASequence(grid_plan=GridRelative(rows=2, columns=2))
     mda5 = MDASequence(
         axis_order="pgc",
         channels=["DAPI", "FITC"],
-        stage_positions=[
-            {
-                "x": 0,
-                "y": 0,
-                "sequence": MDASequence(
-                    grid_plan=GridRelative(overlap=10, rows=2, columns=2)
-                ),
-            }
-        ],
+        stage_positions=[Position(sequence=subseq)],
         keep_shutter_open_across="g",
     )
     for event in mda5:
-        assert event.keep_shutter_open != bool(event.index["g"] == 3)
+        assert event.keep_shutter_open != (event.index["g"] == 3)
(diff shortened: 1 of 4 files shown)


tests/test_sequence.py
```diff
@@ -338,3 +338,49 @@ def test_set_fov_deprecation() -> None:
...
+    mda4 = MDASequence(
+        axis_order="czt",
+        channels=["DAPI", "FITC"],
+        time_plan=TIntervalLoops(loops=2, interval=0),
+        z_plan=ZRangeAround(range=1, step=1),
+        keep_shutter_open_across=("z", "t"),
+    )
+    for event in mda4:
+        is_last_zt = bool(event.index["t"] == 1 and event.index["z"] == 1)
+        assert event.keep_shutter_open != is_last_zt
```
fdrgsp, line 386: Here can we add also this test that uses `keep_shutter_open` in a sequence that also has a sub-sequence? Something like this:

```suggestion
        assert event.keep_shutter_open != is_last_zt

    # testing keep shutter open in sub-sequence
    mda5 = MDASequence(
        axis_order="pgc",
        channels=["DAPI", "FITC"],
        stage_positions=[
            {
                "x": 0,
                "y": 0,
                "sequence": MDASequence(
                    grid_plan=GridRelative(overlap=10, rows=2, columns=2)
                ),
            }
        ],
        keep_shutter_open_across="g",
    )
    for event in mda5:
        assert event.keep_shutter_open != bool(event.index["g"] == 3)
```

fdrgsp (approved):
I like it, better approach than #97.

tests/test_sequence.py
tlambert03: > Here can we add also this test that uses keep_shutter_open in a sequence that also has a sub-sequence? 

nope, the current approach isn't going to work with subsequences (yet).  So [I added this error message if someone tries](https://github.com/pymmcore-plus/useq-schema/pull/119/files#diff-bee8b0e0d2973da762a759ffc19aaea52e9f2be7eb77a40903df6dc0af0622bcR202-R209)

tlambert03: oh... sorry, you mean axes that are only set in the subsequence, yes good idea! will do :)

PR #132, opened 2023-09-01
feat: add generation of random points in grid_plan

fdrgsp:
This PR adds the `RandomPoints` plan, a type of `grid_plan` with the ability to generate a specified number of random points within a specified area.

The random generation is using `numpy` and if a `random_seed` attribute is specified, the generation of points is reproducible.

By setting `allow_overlap=True` together  with `fov_width` and `fov_height`, the generated point will be at least `fov_width` and `fov_height` away from each other (Manhattan distance).

---
For quick test to see how it works:

```py
from useq import RandomPoints
import matplotlib.pyplot as plt
from matplotlib.patches import Ellipse, Rectangle
fig, ax = plt.subplots()
w, h = (10, 10)
fw, fh = (1, 1)
shape = "ellipse"
rn = RandomPoints(
    num_points=10,
    max_width=w,
    max_height=h,
    shape=shape,
    random_seed=0,
    allow_overlap=False,
    fov_width=fw,
    fov_height=fh,
)
if shape == "ellipse":
    ax.add_patch(Ellipse((0, 0), w, h, fill=False, color="m"))
else:
    ax.add_patch(Rectangle((-w / 2, -h / 2), w, h, fill=False, color="m"))
for p in list(rn):
    x, y, _, _, _ = p
    plt.plot(x, y, "go")
    ax.add_patch(Rectangle((x - (fw/2), y - (fh/2)), fw, fh, fill=False,
 color="g"))
plt.axis("equal")
fig.show()
```


src/useq/__init__.py
```diff
@@ -32,6 +38,7 @@
     "AnyZPlan",
     "AutoFocusPlan",
     "AxesBasedAF",
+    "Point",
```
tlambert03, line 41: this looks like extra?

src/useq/_grid.py
```diff
@@ -350,4 +364,194 @@ def _offset_y(self, dy: float) -> float:
...
+
+    return points
+
+
+def _get_iterator(
+    shape: Shape, num_points: int, random_seed: Optional[int]
+) -> Iterator[Tuple[float, ...]]:
+    """Return an iterator of random numbers between 0 and 1 with size depending on the
+    `shape`.
+    """  # noqa: D205
```
tlambert03, line 493: the checks are there for a reason :)  don't use noqa unless it's a very good reason.  these docstrings are not [pep 257 compliant](https://peps.python.org/pep-0257/)

tlambert03 (commented):
a general thought here, (wasn't a good specific place to comment inline in the code)

I love that you're exploring iterators here!  but i think in this case it's gotten a bit more complicated than it needs to be.  here's a few thoughts in no particular order:

1. note that a numpy array is already an iterable, so returning `iter(seed.uniform(0, 1, size=size))` doesn't actually gain anything over just returning `seed.uniform(0, 1, size=size)` (but it is a bit harder to reason about perhaps?)
2. as written, `MAX_ITERS` could more accurately be called `MIN_RANDOM_POINTS`, because when you call `iter_size = max(num_points, MAX_ITER)`, you're guaranteeing that the number of random generated points will be *at least* as bit as `MAX_ITER`, (unless `num_points` is greater).  That's fine in this case, cause numpy is so fast... but the naming might be a bit surprising (even though the number of points does set an upper bound of the maximum possible number of iterations)
3. doing stuff like `math.sqrt(x0) * (max_width / 2) * math.cos(angle * 2 * math.pi)` is going to be slower in a python for loop than it would be if you simply greedily calculated many thousands of candidate points in numpy and iterated over them.
4. What began as a relatively simple pattern (`[p for point in random_points() if _is_a_valid_point(p)] has gotten complicated and passed more parameters through to the deepest functions `_is_a_valid_point` than necessary, and the `PointGenerator` callable type is now very hard to look at and reason about:
    ```python
    PointGenerator = Callable[
        [int, float, float, Tuple[Optional[float], Optional[float]], bool, Optional[int]],
        Iterable[Tuple[float, float]],
    ]
    ```
Note that all shape types share the same `_is_a_valid_point` function... so I think it would be better not to pass `min_distance` and `allow_overlap` through all of those functions... instead just generate your 5000 points up front, and then iterate through _those_ in the main __iter__ function:

so, I'm thinking something like this:

```python
class RandomPoints(_PointsPlan):
    def __iter__(self) -> Iterator[GridPosition]:  # type: ignore
        seed = np.random.RandomState(self.random_seed)
        func = _POINTS_GENERATORS[self.shape]
        points: list[Tuple[float, float]] = []
        for x, y in func(seed, self.max_width, self.max_height):
            if (
                self.allow_overlap
                or (None in (self.fov_width, self.fov_height))
                or _check_validity(self.fov_width, self.fov_height, points, x, y)
            ):
                yield GridPosition(x, y, 0, 0, True)
                points.append((x, y))
            if len(points) >= self.num_points:
                break
```

where:

```python
PointGenerator = Callable[[np.random.RandomState, float, float], np.ndarray]
_POINTS_GENERATORS: dict[Shape, PointGenerator]

def _is_a_valid_point(
    points: list[Tuple[float, float]],
    x: float,
    y: float,
    min_dist_x: float,
    min_dist_y: float,
) -> bool:
    return not any(
        abs(x - point_x) < min_dist_x or abs(y - point_y) < min_dist_y
        for point_x, point_y in points
    )

```

and `_random_points_in_ellipse` and `_random_points_in_rectangle` just greedily make 5000 candidate points and return a numpy array

src/useq/_grid.py
```diff
@@ -350,4 +362,126 @@ def _offset_y(self, dy: float) -> float:
...
+
+def _raise_warning(n_points: int, points: list[Tuple[float, float]]) -> None:
+    """Raise a warning if the number of points is less than the requested number."""
+    warnings.warn(
+        f"Max number of iterations reached ({n_points}). "
+        f"Only {len(points)} points were found.",
+        stacklevel=2,
+    )
+
+
```
tlambert03, line 454: don't make a new function `_raise_warning` just to raise a warning (unless that warning is going to be reused in multiple places)

Also, you don't need to check `if min_dist_x is None or min_dist_y is None` in `_is_a_valid_point` because it already happened in `__iter__`

 
```suggestion
        for x, y in func(seed, n_points, self.max_width, self.max_height):
            if (
                self.allow_overlap
                or self.fov_width is None
                or self.fov_height is None
                or _is_a_valid_point(points, x, y, self.fov_width, self.fov_height)
            ):
                yield GridPosition(x, y, 0, 0, True)
                points.append((x, y))
            if len(points) >= self.num_points:
                break
        if len(points) < self.num_points:
            warnings.warn(
                f"Unable to generate {self.num_points} points non-overlapping points."
                f"Only {len(points)} points were found.",
                stacklevel=2,
            )

    def num_positions(self) -> int:
        return self.num_points


def _is_a_valid_point(
    points: list[Tuple[float, float]],
    x: float,
    y: float,
    min_dist_x: float,
    min_dist_y: float,
) -> bool:
    """Return True if the the point is at least min_dist away from all the others.

    note: using Manhattan distance.
    """
    return not any(
        abs(x - point_x) < min_dist_x and abs(y - point_y) < min_dist_y
        for point_x, point_y in points
    )
```

tlambert03 (approved):
looks good!  All set for merge?

fdrgsp:
> looks good! All set for merge?

yes! Thanks for the help!

PR #145, opened 2023-09-26
feat: make af-device name optional

### LINKED ISSUE #141: Consider making autofocus_device_name optional
I'm wondering whether the [autofocus plan](https://pymmcore-plus.github.io/useq-schema/schema/hardware_autofocus/) has leaked a bit too much implementation details into the model.

I think in the vast majority of cases:
- there will only be one (or no) autofocus device.
- in most cases, I'm guessing that that device will generally be "auto-detectable" by the engine consuming the event (for example, i suspect we could auto-determine what the autofocus device was for micro-manager in the engine for most if not all supported device adapters)

So, that leads me to wonder whether we should update the autofocus plan to be, at minimimum, nothing more than a boolean that says "do autofocus for these axes",  allowing `autofocus_device_name: str | None`.  If None, it would imply "do your best to find it".

This thought comes from looking at https://github.com/pymmcore-plus/pymmcore-widgets/pull/201 and wishing we didn't need the user to enter/select a string device name

tlambert03:
closes #141  by allowing both autofocus device name and offset to be None.
If name is not provided, engines are allowed to simply "apply the offset" as best fits the current hardware (usually, there is only one way to do it at any given time anyway).  If offset is `None`, then the presence of a Hardware Autofocus event simply means "turn on the autofocus without worrying about offset".

see related https://github.com/pymmcore-plus/pymmcore-plus/pull/270


fdrgsp (approved):
Love it!!!

PR #169, opened 2024-07-01
feat: add RelativePosition instead of GridPosition

tlambert03:
when looking at #166, I realized that `GridPosition` is pretty much only there to represent the concept of some "offset" relative to an absolute position.  This PR renames GridPosition to `RelativePosition`, and uses it more braodly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,8 @@ jobs:
     if: always()
     needs: [test]
     uses: pyapp-kit/workflows/.github/workflows/upload-coverage.yml@v2
-    secrets: inherit
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   test-dependents:
     uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@v2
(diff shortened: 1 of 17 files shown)


tlambert03:
note @fdrgsp, this PR removes row/col from `GridPosition`, not for any particular reason, but that i realized it's not actually being used here anywhere... do you have thoughts on where if anywhere those were used?

tlambert03:
found it :) 

```pytb
________________________________________________________________________________
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pymmcore_widgets/_mda/_grid_widget.py", line 554, in _move_to_row_col
    if pos.row == row and pos.col == col:
       ^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pydantic/main.py", line 827, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'RelativePosition' object has no attribute 'row'
```

tlambert03:
i added back row/col ... they're not actually used anywhere, so I'm tempted to remove them, but we can leave that for another discussion.  I like how this unifies the concept of positions used across useq, so would like to get this in before release.  have a look when you can @fdrgsp

PR #171, opened 2024-07-04
feat: few updates for WellPlatePlan

fdrgsp:
@tlambert03 these are few updates I think might be useful. 

- add name to `GridPosition` + update `Position`
- make `well_size` required
- add docstrings

I also added a test to check the serialization because I was getting something weird when working on the `HCSWizard` widget. Basically if you create a `WellPlatePlan` from a `dict`, the `well_points_plan` is always `Position()` no matter what you pass and I don't yet understand why...

### DIFF
--- a/src/useq/_plate.py
+++ b/src/useq/_plate.py
@@ -72,13 +72,15 @@ class WellPlate(FrozenModel):
     columns : int
         The number of columns in the plate.
     well_spacing : tuple[float, float] | float
-        The spacing between wells in the x and y directions.
-        If a single value is provided, it is used for both x and y.
+        The center-to-center distance in mm (pitch) between wells in the x and y
+        directions. If a single value is provided, it is used for both x and y.
     well_size : tuple[float, float] | float
-        The size of each well in the x and y directions.
-        If a single value is provided, it is used for both x and y.
+        The size in mm of each well in the x and y directions. If the well is squared or
+        rectangular, this is the width and height of the well. If the well is circular,
+        this is the diameter. If a single value is provided, it is used for both x and
+        y.
     circular_wells : bool
-        Whether wells are circular (True) or rectangular (False).
+        Whether wells are circular (True) or squared/rectangular (False).
     name : str
         A name for the plate.
     """
@@ -154,6 +156,7 @@ class WellPlatePlan(FrozenModel, Sequence[Position]):
         -   None -> all wells are selected.
         -   0 -> Selects the first row.
         -   [0, 1, 2] -> Selects the first three rows.
+        -   slice(0) -> select no wells
         -   slice(1, 5) -> selects wells from row 1 to row 4.
         -   (2, slice(1, 4)) -> select wells in the second row and only columns 1 to 3.
         -   ([1, 2], [3, 4]) -> select wells in (row, column): (1, 3) and (2, 4)
@@ -378,7 +381,7 @@ class WellPlatePlan(FrozenModel, Sequence[Position]):
 
         return translation @ rotation @ scaling
 
-    def plot(self) -> None:
+    def plot(self, show_axis: bool = True) -> None:
         """Plot the selected positions on the plate."""
         import matplotlib.pyplot as plt
         from matplotlib import patches
@@ -386,7 +389,8 @@ class WellPlatePlan(FrozenModel, Sequence[Position]):
         _, ax = plt.subplots()
 
         # hide axes
-        ax.axis("off")
+        if not show_axis:
+            ax.axis("off")
 
         # ################ draw outline of all wells ################
         height, width = self.plate.well_size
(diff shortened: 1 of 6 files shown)


src/useq/_plate.py
```diff
@@ -63,12 +63,30 @@ def _to_slice(value: Any) -> slice:
...
+        Whether wells are circular (True) or rectangular (False).
+    name : str
+        A name for the plate.
+    """
 
     rows: int
     columns: int
     well_spacing: Tuple[float, float]  # (x, y)
-    well_size: Union[Tuple[float, float], None] = None  # (x, y)
+    well_size: Tuple[float, float]  # (width, height)
```
tlambert03, line 89: my reasoning for making this optional was that it's only necessary if you want to take a few images per well within some area.  It's obviously pretty important in that case... but if you just wanted to take one image per well it's not necessary.  I do guess in that case someone could just put in a fake number if they don't care about where the edge of the well is?  (I do see that it makes life harder elsewhere if you can't rely on it)

fine with me to make it required though

src/useq/_plate.py
```diff
@@ -134,17 +148,24 @@ class WellPlatePlan(FrozenModel, Sequence[Position]):
         "4 rad", "4.5deg").
         If expressed as an arraylike, it is assumed to be a 2x2 rotation matrix
         `[[cos, -sin], [sin, cos]]`, or a 4-tuple `(cos, -sin, sin, cos)`.
+    selected_wells : IndexExpression | None
```
tlambert03, line 151: ❤️ thank you for these :)
could also add `slice(0)` -> select no wells

src/useq/_plate.py
```diff
@@ -165,7 +186,9 @@ def _validate_well_points_plan(
     ) -> Any:
         value = handler(value)
         if plate := info.data.get("plate"):
-            if isinstance(value, RandomPoints):
+            if isinstance(value, RandomPoints) and (
+                value.max_width == np.inf or value.max_height == np.inf
```
tlambert03, line 190: maybe do this on a per-field basis?  this will overwrite both of them if either of them is missing.  Also, could allow shape to pass through if provided (I believe you can check that with `'shape' not in value.__pydantic_fields_set__`

src/useq/_plate.py
```diff
@@ -362,11 +385,11 @@ def plot(self) -> None:
 
         _, ax = plt.subplots()
 
+        # hide axes
+        ax.axis("off")
```
tlambert03, line 389: i left axes in there since it is useful for verifying that the coordinates were actually applied.  otherwise, there's no sense for scale (in microns) and it's hard to be sure that the proper translation/scaling is applied right?

if you want to be able to use this plot elsewhere and don't want those numbers, maybe add a parameter? or  return the axis object so that you could modify it later?

src/useq/_plate.py
```diff
@@ -63,12 +63,30 @@ def _to_slice(value: Any) -> slice:
...
+    """A multi-well plate definition.
+
+    Parameters
+    ----------
+    rows : int
+        The number of rows in the plate.
+    columns : int
+        The number of columns in the plate.
+    well_spacing : tuple[float, float] | float
+        The spacing between wells in the x and y directions.
```
tlambert03, line 75:
```suggestion
        The center-to-center distance (pitch) between wells in the x and y directions.
```

and add units

src/useq/_plate.py
```diff
@@ -63,12 +63,30 @@ def _to_slice(value: Any) -> slice:
...
+        If a single value is provided, it is used for both x and y.
+    well_size : tuple[float, float] | float
+        The size of each well in the x and y directions.
```
tlambert03, line 78: specify whether this is a diameter/radius, edge, etc... and add units

tlambert03 (approved):
nice!  small comments

src/useq/_plate.py
fdrgsp: @tlambert03 do you mean something like that?

```py
@field_validator("well_points_plan", mode="wrap")
@classmethod
def _validate_well_points_plan(
    cls,
    value: Any,
    handler: core_schema.ValidatorFunctionWrapHandler,
    info: core_schema.ValidationInfo,
) -> Any:
    value = handler(value)
    if plate := info.data.get("plate"):
        if isinstance(value, RandomPoints):
            plate = cast(WellPlate, plate)
            kwargs = value.model_dump(mode="python")
            if value.max_width == np.inf:
                kwargs["max_width"] = plate.well_size[0] - (value.fov_width or 0.1)
            if value.max_height == np.inf:
                kwargs["max_height"] = plate.well_size[1] - (value.fov_height or 0.1)
            if 'shape' not in value.__pydantic_fields_set__:
                kwargs["shape"] = (
                    Shape.ELLIPSE if plate.circular_wells else Shape.RECTANGLE
                )
            value = RandomPoints(**kwargs)
    return value
```

tlambert03: i do!

tlambert03 (approved):
Feel free to merge when ready

PR #172, opened 2024-07-05
fix: fix units for correct WellPlatePlan iteration

fdrgsp:
I think we needed to fix the unit of few methods of the `WellPlatePlan` since the `well_points_plan` and the stage coordinates are expressed in `µm` and the `WellPlate` in `mm`.

### DIFF
--- a/src/useq/_plate.py
+++ b/src/useq/_plate.py
@@ -328,7 +328,7 @@ class WellPlatePlan(FrozenModel, Sequence[Position]):
     def all_well_positions(self) -> Sequence[Position]:
         """Return all wells (centers) as Position objects."""
         return [
-            Position(x=x, y=y, name=name)
+            Position(x=x * 1000, y=y * 1000, name=name)  # convert to µm
             for (y, x), name in zip(
                 self.all_well_coordinates, self.all_well_names.reshape(-1)
             )
@@ -338,7 +338,7 @@ class WellPlatePlan(FrozenModel, Sequence[Position]):
     def selected_well_positions(self) -> Sequence[Position]:
         """Return selected wells (centers) as Position objects."""
         return [
-            Position(x=x, y=y, name=name)
+            Position(x=x * 1000, y=y * 1000, name=name)  # convert to µm
             for (y, x), name in zip(
                 self.selected_well_coordinates, self.selected_well_names
             )
@@ -393,6 +393,8 @@ class WellPlatePlan(FrozenModel, Sequence[Position]):
 
         # ################ draw outline of all wells ################
         height, width = self.plate.well_size  # mm
+        height = height * 1000  # µm
+        width = width * 1000  # µm
 
         kwargs = {}
         offset_x, offset_y = 0.0, 0.0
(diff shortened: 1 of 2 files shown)


tlambert03:
good one,  i think i'm gonna play around with using pint here in another PR... should help us avoid this stuff a bit

src/useq/_plate.py
```diff
@@ -356,12 +356,21 @@ def image_positions(self) -> Sequence[Position]:
...
-        # TODO: note that all positions within the same well will currently have the
-        # same name.  This could be improved by modifying `Position.__add__` and
-        # adding a `name` to GridPosition.
-        return [
-            well + offset for well in self.selected_well_positions for offset in offsets
-        ]
+        pos: List[Position] = []
+        for offset in offsets:
+            if isinstance(offset, GridPosition):
+                # invert y axis to use the cartesian coordinate system
```
tlambert03, line 362: i recognize that this is necessary, but it definitely feels like too much local magic: works for your use case at the moment but makes assumptions that may not always be true.  We can do this, but let's open a broader issue about coordinate systems and polarity.

tlambert03 (approved):

src/useq/_plate.py
tlambert03: (actually, this change would have been better for a different PR ... I'm 100% in support of the units changes, but this one gives me pause)

tlambert03: can you remove this, and then open a new issue.  Please come up with a more general proposal for how someone can declare the polarity of their stage coordinate system.  Think about all the places where this matters (I can imagine this comes up with GridRowColumn stuff as well), and then put the solution at a level where it can be accessed by everyone who needs it

fdrgsp: Yes 👍🏻 , what I tried to do within all the HCS logic is to keep the cartesian coordinates constant with y up, -y down, -x left and  +x right.

fdrgsp: @tlambert03 don't you think it is better to just set a coordinate system that useq is using and is up to the user to convert the unit depending on their stage coordinate system? Otherwise we need to specify at least 4 different coordinate systems...

when you say "...come up with a more general proposal for how someone can declare the polarity of their stage coordinate system" are you thinking about something to set the sign of the 4 axis?

tlambert03: yeah, i'm generally observing that this is indeed a sticky point, that I'm still a little bit confused about exactly when and where these coordinate systems need to be declared, what they depend on (is it the stage? is it the plotting system?  is it some transformation system? all of the above? etc...).  And if I'm a bit confused, then I'm certain that the average user would probably be a bit confused...

And so I'd like to see it all given a more formal treatment, some documentation somewhere that
1. explicitly lays out all of the objects in the codebase that are in some way "coordinate-system aware"
2. clarifies who/what defines that coordinate system
3. discusses how it should be customized in certain cases (for example, if my stage treats +X as to the left vs to the right... what should I do)?  

> don't you think it is better to just set a coordinate system that useq is using and is up to the user to convert the unit depending on their stage coordinate system?

once I understand the above a bit better, i think the answer to this question will probably be more obvious.

tlambert03: > what I tried to do within all the HCS logic is to keep the cartesian coordinates constant with y up, -y down, -x left and +x right. Also because we already used this coordinates system for the grid or random points plans for example.

i'm glad you have this clear in your head :) but i don't, and a quick search through the code/documentation didn't reveal this.  I also think it's still potentially unclear what "right" means.  Does it mean "to the right in the image"?  does it mean "the stage moves to the right when I look at it"?  :) reference frames are confusing things and it's hard to be too clear about it all

tlambert03: i'll merge this so you can keep moving forward, but this definitely needs further treatment 👍

PR #174, opened 2024-07-05
use pint for units [very wip]

tlambert03:
this is a proof of principle experiment in using pint.Quantity for various fields.  
I'm not at all convinced it's a good idea.  It's somewhat invasive pint Quantities don't substitute easily for numbers.


tlambert03:
i regret having done this in microsim... so closing this

Issue #173, opened 2024-07-05
TODO: Clarify coordinate systems


tlambert03:
i'm generally observing that this is indeed a sticky point, that I'm still a little bit confused about exactly when and where these coordinate systems need to be declared, what they depend on (is it the stage? is it the plotting system?  is it some transformation system? all of the above? etc...).  And if I'm a bit confused, then I'm certain that the average user would probably be a bit confused...

And so I'd like to see it all given a more formal treatment, some documentation somewhere that
1. explicitly lays out all of the objects in the codebase that are in some way "coordinate-system aware"
2. clarifies who/what defines that coordinate system
3. discusses how it should be customized in certain cases (for example, if my stage treats +X as to the left vs to the right... what should I do)?  

> don't you think it is better to just set a coordinate system that useq is using and is up to the user to convert the unit depending on their stage coordinate system?

once I understand the above a bit better, i think the answer to this question will probably be more obvious.

_Originally posted by @tlambert03 in https://github.com/pymmcore-plus/useq-schema/pull/172#discussion_r1666833839_

fdrgsp:
I spent a bit more time on the useq plate code. As a starting point I think it is easier to define the coordinate system we want  and always use the same. Then maybe we can find a way to let the users set the system they want to use.

We are already using a cartesian coordinate system where -x is left, +x is right, -y is down and +y is up in the `_PointsPlan`. For example, when we create grid points or random points we always assume that the center is (0, 0) and we move along the x and y axis as mention before: -x is left, +x is right, -y is down and +y is up.


<img width="1210" alt="Screenshot 2024-07-05 at 3 51 51 PM" src="https://github.com/pymmcore-plus/useq-schema/assets/70725613/27c93f1a-fd2e-48e4-addf-2c7b5e3ed76d">


So I think it can be a good idea to keep the same coordinate system also for the `WellPlatePlan` for now. This simply means that we need to apply a correct transformation to the set of point that represent the center of the wells of a plate. For example, if we assume that well A1 is centered at (0, 0) and well spacing is 10, then well A2 should 
be centered at (10, 0), well B1 should be centered at (0, -10), and so on. It also means that we do not need to invert any axis when we generate the positions per well since the `_PointsPlan`  use the same coordinate system.
The center well positions looks like this:


[...]

PR #175, opened 2024-07-05
fix: use cartesian coordinate system in WellPlatePlan (-x left, +x right, -y down and +y up)

fdrgsp:
linked to #173 

I removed all the change of sign we had in the `WellPlatePlan` code and only invert the y axis when we apply the transformation so that the y axis is `-y down` and `+y up`. This way we use the same cartesian coordinate system that we are already using in the `_PointsPlans`.

Screenshot 2024-07-05 at 4 15 45 PM


src/useq/_plate.py
```diff
@@ -316,13 +316,15 @@ def selected_well_names(self) -> list[str]:
...
+        """Transform coordinates to the plate coordinate system and invert Y values."""
         # create homogenous coordinates
         h_coords = np.column_stack((coords, np.ones(coords.shape[0])))
         # transform
         transformed = self.affine_transform @ h_coords.T
-        # strip homogenous coordinate
-        return (transformed[:2].T).reshape(coords.shape)  # type: ignore[no-any-return]
+        transformed_coords = (transformed[:2].T).reshape(coords.shape)
+        # invert Y values
+        transformed_coords[:, 0] = -transformed_coords[:, 0]
```
tlambert03, line 326: probably better to do this in the affine transform itself, just invert the scaling factor in the y axis:

```python
scaling[:2, :2] = np.diag([self.plate.well_spacing[0], -self.plate.well_spacing[1]])
```

src/useq/_plate.py
```diff
@@ -357,19 +359,8 @@ def image_positions(self) -> Sequence[Position]:
...
-                    x=offset.x,
-                    y=-offset.y,
-                    row=offset.row,
-                    col=offset.col,
-                    is_relative=offset.is_relative,
-                    name=offset.name,
-                )
-            pos.extend(well + offset for well in self.selected_well_positions)
+        for well in self.selected_well_positions:
+            pos.extend(well + offset for offset in offsets)
```
tlambert03, line 363: 👍

src/useq/_plate.py
fdrgsp: But is it the same? The output won't be identical...

tlambert03: theoretically it should be?  did you try and it wasn't?

tlambert03: yeah, it does work, but the example I gave above was xy, instead of yx...  use this:

```python
        # we invert the Y axis here to go from linearly increasing "index coordinates"
        # to cartesian "plate" coordinates (where y position decreases with increasing index)
        scale_y, scale_x = self.plate.well_spacing
        scaling[:2, :2] = np.diag([-scale_y, scale_x])
```

tlambert03:
please also fix a new scaling issue that we now have in random points:

```python
                if value.max_width == np.inf:
                    kwargs["max_width"] = plate.well_size[0] - (value.fov_width or 0.1)
                if value.max_height == np.inf:
                    kwargs["max_height"] = plate.well_size[1] - (
                        value.fov_height or 0.1
                    )
```

the `plate.well_sizes` need to be scaled by 1000

src/useq/_plate.py
fdrgsp: yup, I tried that as well but then the test give a slightly different result and if you use a rotation and plot the plate it is rotating in the opposite direction...

tlambert03: tell me what specific differences you're seeing

src/useq/_plate.py
```diff
@@ -368,18 +366,24 @@ def affine_transform(self) -> np.ndarray:
         """Return transformation matrix.
 
         This includes:
-        1. scaling by plate.well_spacing
+        1. translation to a1_center_xy
         2. rotation by rotation_matrix
-        3. translation to a1_center_xy
+        3. scaling by plate.well_spacing
```
tlambert03, line 371: actually these should stay in the order they were.  when you compose transforms (like `translation @ rotation @ scaling`) it goes from right to left

src/useq/_plate.py
fdrgsp: Ok I think I was not convinced for another reason...I was a bit confused with the output numbers of the positions. But I think it was because we had to also consider the units of  `a1_center_xy`. I think we can say that `a1_center_xy` is in `µm` since most likely the stage coords are in µm (and the `well_points_plan` too) and we covert it during translation.

src/useq/_plate.py
fdrgsp: yup sorry

src/useq/_plate.py
tlambert03: yeah, it's confusing!  Should we consider expressing well sizes in microns too just to keep everything consistent?  Is this the only place in useq where we use mm?

fdrgsp: Yes I believe is the only place we use mm...you're right, maybe it is worth using µm for everything at this point...

tlambert03: i know that's also problematic since all plates specs will be in mm... maybe skip that for now

fdrgsp: then I think this is ready for review

fdrgsp:
@tlambert03 out of curiosity, if you run the `test` for the plate locally, does `test_plate_plan_serialization` fail for you?

tlambert03:
nope, and not on CI either obviously... whats your traceback

fdrgsp:
> nope, and not on CI either obviously... whats your traceback

the well_points_plan is always defaulting to Position()...
```
    def test_plate_plan_serialization() -> None:
        pp = useq.WellPlatePlan(
            plate=96,
            a1_center_xy=(500, 200),
            rotation=5,
            selected_wells=np.s_[1:5:2, :6:3],
            well_points_plan=useq.RandomPoints(num_points=10),
        )
        js = pp.model_dump_json()
        pp2 = useq.WellPlatePlan.model_validate_json(js)
>       assert pp2 == pp
E       AssertionError: assert WellPlatePlan(plate=WellPlate(rows=8, columns=12, well_spacing=(9.0, 9.0), well_size=(6.4, 6.4)), a1_center_xy=(500.0, 200.0), rotation=5.0, selected_wells=(slice(1, 5, 2), slice(None, 6, 3)), well_points_plan=Position()) == WellPlatePlan(plate=WellPlate(rows=8, columns=12, well_spacing=(9.0, 9.0), well_size=(6.4, 6.4)), a1_center_xy=(500.0, 200.0), rotation=5.0, selected_wells=(slice(1, 5, 2), slice(None, 6, 3)), well_points_plan=RandomPoints(num_points=10, max_width=6399.9, max_height=6399.9))
E
E         At index 0 diff: Position(name='B1') != Position(x=2010.7489791050302, y=-10177.428033516471, name='B1_0000')
E         Right contains 36 more items, first extra item: Position(x=1776.901041996079, y=-9937.020701044301, name='B1_0004')
E
E         Full diff:
E         - WellPlatePlan(plate=WellPlate(rows=8, columns=12, well_spacing=(9.0, 9.0), well_size=(6.4, 6.4)), a1_center_xy=(500.0, 200.0), rotation=5.0, selected_wells=(slice(1, 5, 2), slice(None, 6, 3)), well_points_plan=RandomPoints(num_points=10, max_width=6399.9, max_height=6399.9))
E         ?                                                                                                                                                                                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E         + WellPlatePlan(plate=WellPlate(rows=8, columns=12, well_spacing=(9.0, 9.0), well_size=(6.4, 6.4)), a1_center_xy=(500.0, 200.0), rotation=5.0, selected_wells=(slice(1, 5, 2), slice(None, 6, 3)), well_points_plan=Position())
E         ?                                                                                                                                                                                                                   ^^^^^^^^^^^
```

tlambert03:
> the well_points_plan is always defaulting to Position()...

strange... not sure.  what's in your env?
it must be environment, since it's working on CI, but I wonder what it could be.

src/useq/_plate.py
```diff
@@ -316,7 +316,7 @@ def selected_well_names(self) -> list[str]:
         return list(self.all_well_names[self.selected_wells].reshape(-1))
 
     def _transorm_coords(self, coords: np.ndarray) -> np.ndarray:
-        """Transform coordinates to the plate coordinate system."""
+        """Transform coordinates to the plate coordinate system and invert Y values."""
```
tlambert03, line 319:
```suggestion
        """Transform coordinates to the plate coordinate system."""
```

fdrgsp:
> > the well_points_plan is always defaulting to Position()...
> 
> strange... not sure. what's in your env? it must be environment, since it's working on CI, but I wonder what it could be.

ok fixed it, not sure what was the conflict...

tlambert03 (approved):
good stuff, thanks

PR #177, opened 2024-07-08
feat: formalizing point-visiting strategies

tlambert03:
adds 

- useq.TraversalOrder - to dictate how random points are traversed
- addes a nearest neighbor and  2-opt algorithm, (for slightly better path choosing)
- allows you to specify the starting index
- adds plotting to all points plans:

```python
import useq

pp = useq.RandomPoints(
    num_points=100,
    max_height=6000,
    max_width=6000,
    order=useq.TraversalOrder.TWO_OPT,
    random_seed=1,
    shape="rectangle",
    start_at=10,
    fov_height=300,
    fov_width=300,
    allow_overlap=False,
)
pp.plot()

```

two-opt:

Screenshot 2024-07-08 at 4 01 09 PM

nearest neighbor:
Screenshot 2024-07-08 at 4 01 34 PM

### DIFF
--- a/src/useq/__init__.py
+++ b/src/useq/__init__.py
@@ -19,7 +19,7 @@ from useq._mda_event import MDAEvent, PropertyTuple
 from useq._mda_sequence import MDASequence
 from useq._plate import WellPlate, WellPlatePlan
 from useq._plate_registry import register_well_plates, registered_well_plate_keys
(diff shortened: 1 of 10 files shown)


tlambert03:
this is going to temporarily break one of the new point plan selectors in pymmcore-widgets, because requires a max_width > 0

Issue #187, opened 2024-09-19
Position plan timing resets for each position


wl-stepp:
I will open a new issue for this, but it's very similar to #149

On useq-schema-0.4.7:
```
>>> import useq                                                                                       
>>> seq = useq.MDASequence(stage_positions=[(100,100), (0,0)], time_plan={"interval": 1, "loops": 2}, axis_order='ptgcz')
>>> list(seq)
[MDAEvent(index=mappingproxy({'t': 0, 'p': 0}), min_start_time=0.0, x_pos=100.0, y_pos=100.0), 
MDAEvent(index=mappingproxy({'t': 1, 'p': 0}), min_start_time=1.0, x_pos=100.0, y_pos=100.0), 
MDAEvent(index=mappingproxy({'t': 0, 'p': 1}), min_start_time=0.0, x_pos=0.0, y_pos=0.0), 
MDAEvent(index=mappingproxy({'t': 1, 'p': 1}), min_start_time=1.0, x_pos=0.0, y_pos=0.0)]
```
See how the min_start_time resets.
With this the first position runs fine, but the later ones will run at full speed.

wl-stepp:
I think the problem with integration in pymmcore-plus might be a bit deeper. For fast acquisitions on larger samples, this could still be a problem even if the counter would not reset. Maybe the timer in pymmcore-plus should actually reset for a new position after the position is set? Might be worth thinking about a little more.

tlambert03:
@wl-stepp, on the useq-schema side, what do you think the `MDAEvent` object *should* look like in this case?  the most straightforward correction here could be to just offset p1 to have a min_start time at least as high as the last event in the previous position:

```python
[MDAEvent(index=mappingproxy({'t': 0, 'p': 0}), min_start_time=0.0, x_pos=100.0, y_pos=100.0), 
MDAEvent(index=mappingproxy({'t': 1, 'p': 0}), min_start_time=1.0, x_pos=100.0, y_pos=100.0), 
MDAEvent(index=mappingproxy({'t': 0, 'p': 1}), min_start_time=1.0, x_pos=0.0, y_pos=0.0), 
MDAEvent(index=mappingproxy({'t': 1, 'p': 1}), min_start_time=2.0, x_pos=0.0, y_pos=0.0)]
```

That's *slightly* better, but would still be insufficient in terms of actually yielding the desired result. 

We could could also allow an event to "reset" t0:

```python
[...]

PR #194, opened 2024-11-01
fix: reset_event_timer defaults to True in sub-sequences

wl-stepp:
Fixes problem mentioned in #187. All events in sub-sequences had the reset_event_timer flag set to True.
https://github.com/pymmcore-plus/useq-schema/issues/187#issuecomment-2450480509

I'm not sure if this is the correct position for this fix, please check if this makes sense.


tlambert03 (approved):
this looks great!  Thanks for catching and fixing @wl-stepp

PR #230, opened 2025-05-12
tests: use setup-mm action

tlambert03:
trying the new https://github.com/pymmcore-plus/setup-mm-test-adapters


tlambert03:
hooray!  50s down to 20s

Issue #256, opened 2026-02-26
Add xy_stage and z_stage fields to Position and MDAEvent


ieivanov:
## Problem

Currently, `Position` only defines `x`, `y`, `z` coordinates. When the engine moves to a position, it always uses the default XY stage (`core.getXYStageDevice()`) and default focus device (`core.getFocusDevice()`). Some microscope setups have multiple stage devices, and users need to target specific stages during acquisition. Explicitly specifying the XY and Z stages to be used during acquisition would make the acquisition more reliable — as it no longer relies on the previous state of `core.getXYStageDevice()` and `core.getFocusDevice()` — and serve to more precisely document the acquisition protocol through an MDASequence yaml file.

### useq-schema changes

1. **`PositionBase`** (`src/useq/_position.py`): Add `xy_stage: str | None = None` and `z_stage: str | None = None`. Defaults to `None` (use the default device), preserving backward compatibility.

2. **`MDAEvent`** (`src/useq/_mda_event.py`): Add matching `xy_stage: str | None = None` and `z_stage: str | None = None` fields, plus entries in `MDAEvent.Kwargs`.

3. **Event iteration** (`src/useq/_iter_sequence.py`): Propagate `xy_stage` and `z_stage` from the Position through `_xyzpos()` into the event dict.

### pymmcore-plus engine changes

In `src/pymmcore_plus/mda/_engine.py`:

- `_set_event_xy_position()`: If `event.xy_stage` is provided, call `core.setXYPosition(xy_stage, x, y)` and use the stage-specific key for last-position caching. Otherwise, current behavior (default device).
- `_set_event_z()`: If `event.z_stage` is provided, call `core.setPosition(z_stage, z)` instead of `core.setZPosition(z)`. Otherwise, current behavior.
- `setup_sequenced_event()`: Use `event.xy_stage or core.getXYStageDevice()` and `event.z_stage or core.getFocusDevice()` for stage sequence calls.

## Note on useq v2

Some or all of these needs may be addressed in a future `useq.v2` schema revision. I don't currently have clarity on when that schema will be implemented in the engine, so this proposal targets the current v1 API for near-term use.

@tlambert03 I welcome your input here before we work on an implementation.

tlambert03:
closed by #266

PR #261, opened 2026-02-28
feat: Add ROI to MDAEvent and setup to MDASequence

ieivanov:
See https://github.com/pymmcore-plus/useq-schema/issues/260

Closes #258 
Closes #259

### DIFF
--- a/src/useq/_mda_sequence.py
+++ b/src/useq/_mda_sequence.py
@@ -70,8 +70,8 @@ class MDASequence(UseqModel):
         An optional setup event to prepend to the sequence. This event is yielded
         first during iteration and can be used to set hardware state (e.g. device
         properties, ROI) before the main acquisition events. Typically used with
-        `action=CustomAction(name="setup")` so the engine applies hardware state
-        without acquiring an image. By default, `None`.
+        `action=SetupAction()` so the engine applies hardware state without
+        acquiring an image. By default, `None`.
     keep_shutter_open_across : tuple[str, ...]
         A tuple of axes `str` across which the illumination shutter should be kept open.
         Resulting events will have `keep_shutter_open` set to `True` if and only if
@@ -180,6 +180,27 @@ class MDASequence(UseqModel):
        range: 3.0
        step: 1.0
     ```
+
+    Create a sequence with a setup event
+
+    >>> from useq import MDASequence, MDAEvent, SetupAction
+    >>> seq = MDASequence(
+    ...     setup=MDAEvent(
+    ...         properties=[("Camera", "Binning", "2")],
+    ...         roi=(0, 0, 512, 512),
+    ...         action=SetupAction(),
+    ...     ),
+    ...     channels=["GFP"],
+    ...     time_plan={"interval": 1, "loops": 5},
+    ... )
+
+    The setup event is the first event yielded during iteration:
+
+    >>> events = list(seq)
+    >>> events[0].action.type
+    'setup'
+    >>> events[0].roi
+    (0, 0, 512, 512)
     """
 
     metadata: dict[str, Any] = Field(default_factory=dict)
--- a/src/useq/_mda_sequence.py
+++ b/src/useq/_mda_sequence.py
@@ -70,8 +70,8 @@ class MDASequence(UseqModel):
         An optional setup event to prepend to the sequence. This event is yielded
         first during iteration and can be used to set hardware state (e.g. device
         properties, ROI) before the main acquisition events. Typically used with
-        `action=SetupAction()` so the engine applies hardware state without
-        acquiring an image. By default, `None`.
+        `action=CustomAction(name="setup")` so the engine applies hardware state
+        without acquiring an image. By default, `None`.
     keep_shutter_open_across : tuple[str, ...]
         A tuple of axes `str` across which the illumination shutter should be kept open.
         Resulting events will have `keep_shutter_open` set to `True` if and only if
@@ -183,12 +183,12 @@ class MDASequence(UseqModel):
 
     Create a sequence with a setup event
 
-    >>> from useq import MDASequence, MDAEvent, SetupAction
+    >>> from useq import MDASequence, MDAEvent, CustomAction
     >>> seq = MDASequence(
     ...     setup=MDAEvent(
     ...         properties=[("Camera", "Binning", "2")],
     ...         roi=(0, 0, 512, 512),
-    ...         action=SetupAction(),
+    ...         action=CustomAction(name="setup"),
     ...     ),
     ...     channels=["GFP"],
     ...     time_plan={"interval": 1, "loops": 5},
@@ -197,10 +197,32 @@ class MDASequence(UseqModel):
     The setup event is the first event yielded during iteration:
 
     >>> events = list(seq)
-    >>> events[0].action.type
+    >>> events[0].action.name
     'setup'
     >>> events[0].roi
     (0, 0, 512, 512)
+
+    Print setup sequence as yaml
+
+    >>> print(seq.yaml())
+    channels:
+    - config: GFP
+    setup:
+      action:
+        name: setup
+      properties:
+      - - Camera
+        - Binning
+        - '2'
+      roi:
+      - 0
+      - 0
+      - 512
+      - 512
+    time_plan:
+      interval: 1.0
+      loops: 5
+    <BLANKLINE>
     """
 
     metadata: dict[str, Any] = Field(default_factory=dict)
--- a/src/useq/_mda_sequence.py
+++ b/src/useq/_mda_sequence.py
@@ -66,6 +66,12 @@ class MDASequence(UseqModel):
         generated, do not set.
     autofocus_plan : AxesBasedAF | None
         The hardware autofocus plan to follow. One of `AxesBasedAF` or `None`.
+    setup : MDAEvent | None
+        An optional setup event to prepend to the sequence. This event is yielded
+        first during iteration and can be used to set hardware state (e.g. device
+        properties, ROI) before the main acquisition events. Typically used with
+        `action=CustomAction(name="setup")` so the engine applies hardware state
+        without acquiring an image. By default, `None`.
     keep_shutter_open_across : tuple[str, ...]
         A tuple of axes `str` across which the illumination shutter should be kept open.
         Resulting events will have `keep_shutter_open` set to `True` if and only if
@@ -188,6 +194,7 @@ class MDASequence(UseqModel):
     time_plan: AnyTimePlan | None = None
     z_plan: AnyZPlan | None = None
     autofocus_plan: AnyAutofocusPlan | None = None
+    setup: MDAEvent | None = None
     keep_shutter_open_across: tuple[str, ...] = Field(default_factory=tuple)
 
     _uid: UUID = PrivateAttr(default_factory=uuid4)
--- a/src/useq/_mda_sequence.py
+++ b/src/useq/_mda_sequence.py
@@ -200,7 +200,7 @@ class MDASequence(UseqModel):
     >>> events[0].action.name
     'setup'
     >>> events[0].roi
-    CameraROI(x=0, y=0, width=512, height=512)
+    CameraROI(offset_x=0, offset_y=0, width=512, height=512)
 
     Print setup sequence as yaml
 
@@ -216,9 +216,9 @@ class MDASequence(UseqModel):
         - '2'
       roi:
         height: 512
+        offset_x: 0
+        offset_y: 0
         width: 512
-        x: 0
-        y: 0
     time_plan:
       interval: 1.0
       loops: 5
--- a/src/useq/_mda_sequence.py
+++ b/src/useq/_mda_sequence.py
@@ -70,9 +70,7 @@ class MDASequence(UseqModel):
     setup : MDAEvent | None
         An optional setup event to prepend to the sequence. This event is yielded
         first during iteration and can be used to set hardware state (e.g. device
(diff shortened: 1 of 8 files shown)


src/useq/_mda_sequence.py
```diff
@@ -188,6 +237,7 @@ class MDASequence(UseqModel):
     time_plan: AnyTimePlan | None = None
     z_plan: AnyZPlan | None = None
     autofocus_plan: AnyAutofocusPlan | None = None
+    setup: MDAEvent | None = None
```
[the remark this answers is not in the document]
tlambert03: hey @ieivanov, any thoughts on this comment?  this is the main thing i'm wanting to see resolved before merging this.  Want to make sure we're not adding something that _requires_ manual boilerplate (every time) in order to avoid unexpected behavior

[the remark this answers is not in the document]
ieivanov: Sorry for the delay - I'm all for reducing the boilerplate code. I agree that having to specify 

```python
setup=MDAEvent(
    ...
    action=CustomAction(name="setup"),
)
```

every time is a bit tedious and prone to errors as you suggested. I'm open to different implementations on how to address this. To me, the most straightforward would be a field validator in MDASequence - let me know what you think.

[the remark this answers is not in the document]
tlambert03: > a field validator in MDASequence

yep!  sounds good.  I think it should:
- if `action` is None... set it to `CustomAction(name="setup")`
- if `action` is `AcquireImage` ...  I would actually be happy with raising a `ValueError`, what do you think?
- if `action` is anything else I think we should leave it.

I suppose an after-validator is appropriate here.  if you run into `MDAEvent` being immutable issues, go ahead and use `object.__setattr__`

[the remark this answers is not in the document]
ieivanov: Let me know what you think of [a51dd1a](https://github.com/pymmcore-plus/useq-schema/pull/261/commits/a51dd1ab624cc7e41381e99eca7f7f1bb37eafd0). I needed one workaround since `MDAEvent` defaults to `action=AcquireImage`

ieivanov:
I updated `MDAEvent.roi` to be a `CameraROI` instead of a tuple, but now I'm having second thoughts. Should we assume that the `MDAEvent` will use the default Camera, FocusDevice, SLM, etc. or does it make sense to provide infrastructure to change these per `MDAEvent`? I think the former makes more sense, I don't have strong use cases for the latter yet. In that case, it's better that `MDAEvent.roi` is a tuple that will be applied to `core.getCameraDevice()`.

The only time this gets a bit more complicated is when it comes to positions. `MDAEvent.z_pos` refers to `core.getFocusDevice()`, for example in `setup_sequenced_event` we call `if event.z_sequence: core.startStageSequence(core.getFocusDevice())`. I think it's also OK for `x_pos` and `y_pos` to also refer to `core.getXYStageDevice()`. The only strong need I currently have is to extend the definition of Positions to include stages other than X, Y, and Z, in which case these stages must be associated with a device label.

This is probably more of a pymmcore-plus discussion rather than a useq-schema discussion.

tlambert03:
in general, I prefer structured objects with names, rather than tuples with positional meaning.  So yes, I much prefer `CameraRoi` to `tuple[int, int, int, int]`  

> Should we assume that the MDAEvent will use the default Camera, FocusDevice, SLM, etc. or does it make sense to provide infrastructure to change these per MDAEvent

well... this really is the first *explicit* camera-specific setting (outside of exposure, which has been a channel specific setting, which actually still allows for different cameras depending on the config object).  And one main reason why in https://github.com/pymmcore-plus/useq-schema/issues/260#issuecomment-3969955454 i was proposing that we use `rois: list[CameraRoi]`

> The only time this gets a bit more complicated is when it comes to positions. `MDAEvent.z_pos` refers to `core.getFocusDevice()`

I  wouldn't state it this way.  I would say that `MDAEngine` *currently interprets it this way*.  You should not look at MDAEngine's implementation as the source of truth for what useq schema fields mean.  but also... I'm not quite sure what the relevance is here for the ROI discussion?    Let's definitely leave the stage device discussion for a different PR, so as not to confuse things here.

let's just focus on "correctness for setting an ROI associated with an event".  as mentioned in that previous comment.  I think `list[CameraRoi]` is totally future proof and unambiguous, at the risk of being slightly confusing since it would be the only place where multiple camera devices are explicitly mentioned.  So I'm ok with `roi: CameraRoi | None` as well (and if we change it to a list in the future, it's not a breaking change to make a new attribute called `rois` instead of `roi`)

tlambert03:
> I think the former makes more sense, I don't have strong use cases for the latter yet. In that case, it's better that MDAEvent.roi is a tuple that will be applied to core.getCameraDevice().

just to specifically respond to this:  I don't see the need for the camera device label as the thing determining whether we should use a `tuple[int,int,int,int]` vs a model with four fields.  I think we *should* use a model with four fields, regardless of whether we choose to add a `label` field to that model or not.  And, if you want to, you could add a `model_validator(mode='before')` that casts a plan tuple to a dict with those four fields, for convenience in the constructor

ieivanov:
> I don't see the need for the camera device label as the thing determining whether we should use a tuple[int,int,int,int] vs a model with four fields. I think we should use a model with four fields, regardless of whether we choose to add a label field to that model or not.

Sorry, I didn't mean to argue for using a tuple over a model. I do also like 

```python
class CameraROI(UseqModel):
    x: int
    y: int
    width: int
    height: int
```

My hesitation was whether or not we should include the camera `label`. This will be only the second time we introduce the concept of a "device" in `MDAEvent` (apart from `SLMImage.device`) - and this time I think it's a bit more consequential, as - at least to me, it's not perfectly clear how it should be handled. I think there are three options.

## Option 1

Both exposure time and ROI are properties of the camera device. We could have 

```python
class CameraROI(UseqModel):
    x: int
    y: int
    width: int
    height: int
    camera: str: None = None
    
class CameraExposure(UseqModel):
    exposure_time: float
    camera: str | None = None

class MDAEvent:
    ...
    roi: CameraROI | None = None
    exposure: CameraExposure | None = None
```

(this will be a breaking change, echoing our discussion of StagePosition https://github.com/pymmcore-plus/useq-schema/issues/260#issuecomment-3969929587)

If we do decide on this concept - properties like ROI and exposure are associated with a specific camera device, should we check that in a given `MDAEvent` the ROI and exposure reference the same camera? In the event is an `AcquireImage` event, should we assume that the engine will acquire an image using that same camera? Are there cases where different devices for `CameraROI`, tentative `CameraExposure`, and image acquisition make sense?

## Option 2

`camera` is a property of MDAEvent, as in

```python
class CameraROI(UseqModel):
    x: int
    y: int
    width: int
    height: int

class MDAEvent:
    ...
    roi: CameraROI | None = None
    exposure: float | None = None
    camera: str | None = None
```

If `MDAEvent.camera` is provided we would expect that this camera is used for acquisition - in an acquisition engine we'd call `core.setCameraDevice(MDAEvent.camera)`. ROI and exposure will be applied afterwards using `core.setROI()`, i.e. without referencing a specific camera. If `MDAEvent.camera = None` will will not call `core.setCameraDevice` - the event will use a pre-configured `core.getCameraDevice()` or one set through the channel config.

## Option 3

We don't include references to specific devices in `MDAEvent` - we'd expect that these are configured before the event is executed. In this case we'd have

 ```python
class CameraROI(UseqModel):
    x: int
    y: int
    width: int
    height: int

class MDAEvent:
    ...
    roi: CameraROI | None = None
    exposure: float | None = None
#  camera: str | None = None, i.e. no camera device is declared
```

If desired, these can be configured using

```python
MDASequence(
    setup=MDAEvent(
        properties=[("Core", "Camera", "camera_name")]
    )
)
```

We can then directly call `core.setROI()` and `core.setExposure` without specifying a camera label.

___
In Options 2 and 3 it may be good to rethink how the SLM device is defined and used for consistency, but not critical, I think.

On the issue of one camera versus multiple cameras. My preference would be for a single `roi: CameraRoi | None`. As you said, in the future we can extend to `rois` and `exposures` (multiple cameras don't need to have the same exposure time, we often use the multicamera adapter to acquire brightfield, GFP and RFP images simultaneously all of which use different exposure time).

Let me know what you think!

tlambert03:
ah, sorry i misunderstood.  I see now. 

Of these options, I think I like 3 the best as the one that gets us the feature we need with the least fuss.

But I do like where you're going with the label options.  There's a fourth option I don't see yet which is to just combine all camera-specific things (anything that unambiguously refers to a single camera) into one new `camera` field (and provide an exposure `@property` for backwards compatibility):

```python
class CameraRoi(UseqModel):
    roi_x: NonNegativeInt
    roi_y: NonNegativeInt
    roi_width: NonNegativeInt
    roi_height: NonNegativeInt

class Camera(UseqModel):
    roi: CameraRoi | None = None
    exposure: NonNegativeFloat | None = None
    binning: PositiveInt | None = None
    device: str | None = None

class MDAEvent(UseqModel):
    ...
    camera: Camera | None = None

    @property
    def exposure(self) -> float | None:
        if self.camera is not None:
            return self.camera.exposure
        return None

    @exposure.setter
    def exposure(self, value: float | None) -> None:
        if self.camera is None:
            self.camera = Camera(exposure=value)
        else:
            self.camera.exposure = value
```

do you have a preference between option 3 and that?  I'm happy either way

ieivanov:
>There's a fourth option I don't see yet which is to just combine all camera-specific things (anything that unambiguously refers to a single camera) into one new camera field (and provide an exposure @property for backwards compatibility):

I think this is a good structure too! 

For the purposes of this PR, I'm inclined to go with Option 3 - like you said, it gets the job done and is least amount of fuss. I think we'll end up revising the discussion of specifying device labels in `MDAEvent` when we tackle #256 and #257. Let's pick this up them? I think in that context we'll have stronger use cases.

ieivanov:
Oook, I think I'm happy with this PR, @tlambert03 let me know what you think.

src/useq/_mda_sequence.py
```diff
@@ -66,6 +67,12 @@ class MDASequence(UseqModel):
         generated, do not set.
     autofocus_plan : AxesBasedAF | None
         The hardware autofocus plan to follow. One of `AxesBasedAF` or `None`.
+    setup : MDAEvent | None
+        An optional setup event to prepend to the sequence. This event is yielded
+        first during iteration and can be used to set hardware state (e.g. device
+        properties, ROI) before the main acquisition events. Typically used with
+        `action=CustomAction(name="setup")` so the engine applies hardware state
+        without acquiring an image. By default, `None`.
```
tlambert03:
```suggestion
        properties, ROI) before the main acquisition events. By default, `None`.
```

what do you think?  maybe hide that detail since we're enforcing it now?

src/useq/_mda_sequence.py
```diff
@@ -205,6 +255,31 @@ def __hash__(self) -> int:
...
+    def _validate_setup(cls, v: Any) -> Any:
+        if v is None:
+            return v
+        from useq._mda_event import MDAEvent
+
+        if isinstance(v, MDAEvent):
+            if isinstance(v.action, AcquireImage):
+                # AcquireImage is the default action for MDAEvent; replace it
+                # with a CustomAction to avoid unintended image acquisition.
+                v = v.model_copy(update={"action": CustomAction(name="setup")})
```
tlambert03: hmmm,  yeah: `AcquireImage` being the default is definitely annoying here.  The difference between this path (which silently overrides) and the one below (which raises an exception) is problematic though.  Specifically, if the user *explicitly* used AcquireImage, we silently overwrite it:

```python
In [19]: MDASequence(setup=MDAEvent(action=AcquireImage()))
Out[19]: MDASequence(setup=MDAEvent(action=CustomAction(name='setup')))
```

The trick here is to use pydantic's `model_fields_set` to determine whether the `action` was actually supplied by the user, or just the default:

```python
In [20]: MDAEvent().model_fields_set
Out[20]: set()

In [21]: MDAEvent(action=AcquireImage()).model_fields_set
Out[21]: {'action'}
```

so after `if isinstance(v.action, AcquireImage):`, I think you should check`'action' in v.model_fields_set` and raise the same value error

src/useq/_mda_sequence.py
ieivanov: Yup, thanks!

src/useq/_mda_sequence.py
ieivanov: Something like this: https://github.com/pymmcore-plus/useq-schema/pull/261/changes/fb5ab71c9efa72759ed3ea58104aacea493014a3?

tlambert03: thanks, yes that's exactly what I meant.  

after looking at it though, I realized this would be a better candidate for an after validator (`mode='after'`) ... so you don't need to deal with all the possible pre-cast types (you *only* need to check whether it was explicitly set to `AcquireImage`)

I just made the change and pushed it so you don't have to keep making changes :) 

```python
    @field_validator("setup", mode="after")
    def _validate_setup(cls, v: MDAEvent | None) -> MDAEvent | None:
        if v is None:
            return v
        if isinstance(v.action, AcquireImage):
            if "action" in v.model_fields_set:
                raise ValueError(
                    "Setup event action cannot be 'AcquireImage'. "
                    "Omit the action field to automatically use "
                    "CustomAction(name='setup'), or use a different "
                    "action type."
                )
            v = v.model_copy(update={"action": CustomAction(name="setup")})
        return v
```

tlambert03 (approved):
perfect!  That's for tolerating all the discussion.  Merging now and we can move on to thinking about stages

Issue #196, opened 2024-11-01
Feat:  possible add non-dimension coordinates to the MDAEvent


tlambert03:
Many libraries like xarray, netCDF, pandas, and GDAL have the concept of [dimension coordinates and non-dimension coordinates](https://docs.xarray.dev/en/latest/user-guide/data-structures.html#coordinates), where dimension coords are used for label based indexing and alignment and non-dimension coords provide additional info that can be useful for plotting, etc...

when we iterate over grid positions, the `MDAEvent.index` attribute only gets a single `p` index, but it would be useful to also pass on row/column, etc...  The same concept could arise in other cases, and it would be nice to have a place for that.  Could be an additional field, or some sort of convention in index (like `{'p': 4, '_r': 1, '_c': 1}`

PR #266, opened 2026-03-15
feat: add Position.properties to be yielded on the first index of each position

tlambert03:
@ieivanov I'm wondering if this would work for you to close the remainder of #260 (specifically, the case of multi-actuator positioning).  `useq.Position` gains a `properties` field.  usage when setting up a useq MDASequence would look like this:

```python
import useq

seq = useq.MDASequence(
    # core FocusDevices should be set to fast (e.g. galvo/piezo) device.
    setup=MDAEvent(properties=[("Core", "Focus", "PiezoZ")],),
    stage_positions=[
        # ObjectiveZ is the name of your slow z-positioner (e.g. nosepiece motor)
        useq.Position(x=100, y=200, properties=[("ObjectiveZ", "Position", 50.0)]),
        useq.Position(x=300, y=400, properties=[("ObjectiveZ", "Position", 75.0)]),
    ],
    z_plan=useq.ZRangeAround(range=4, step=2),
)
```

the semantics are that each time a position is newly visited, the properties are included in the event:

```python
>>> list(seq)

[
    MDAEvent(index={'p': 0, 'z': 0}, x_pos=100.0, y_pos=200.0, z_pos=-2.0, properties=[PropertyTuple(device_name='ObjectiveZ', property_name='Position', property_value=50.0)]),
    MDAEvent(index={'p': 0, 'z': 1}, x_pos=100.0, y_pos=200.0, z_pos=0.0),
    MDAEvent(index={'p': 0, 'z': 2}, x_pos=100.0, y_pos=200.0, z_pos=2.0),
    MDAEvent(index={'p': 1, 'z': 0}, x_pos=300.0, y_pos=400.0, z_pos=-2.0, properties=[PropertyTuple(device_name='ObjectiveZ', property_name='Position', property_value=75.0)]),
    MDAEvent(index={'p': 1, 'z': 1}, x_pos=300.0, y_pos=400.0, z_pos=0.0),
    MDAEvent(index={'p': 1, 'z': 2}, x_pos=300.0, y_pos=400.0, z_pos=2.0)
]
```

That's a *choice*.  We could also choose to include the props in *every* event yielded by that position, and then it would be up to the Engine to track the diff and not make unnecessary changes (so, a bit simpler to have it done here in useq if that's ok).

on the pymmcore plus side, support for this would look check whether `property_name == Keyword.Position` and treat it specially

```python
    def _set_event_properties(self, properties: Sequence[useq.PropertyTuple]) -> None:
        """Set device properties, using setPosition for stage devices."""
        core = self.mmcore
        for dev, prop, value in properties:
            try:
                if prop == Keyword.Position:
                    dev_type = core.getDeviceType(dev)
                    if dev_type == DeviceType.XYStage:
                        if not isinstance(value, (list, tuple)):
                            logger.warning(
                                "Expected XY position to be a list or tuple, got %s. "
                                "Ignoring.",
                                type(value),
                            )
                            continue
                        x, y, *_ = value
                        core.setXYPosition(dev, float(x), float(y))
                    elif dev_type == DeviceType.Stage:
                        core.setPosition(dev, float(value))
                    else:
                        core.setProperty(dev, prop, value)
                else:
                    core.setProperty(dev, prop, value)
            except Exception as e:
                logger.warning(
                    "Failed to set property %s of device %s. %s", prop, dev, e
                )
```

### DIFF
--- a/src/useq/_iter_sequence.py
+++ b/src/useq/_iter_sequence.py
@@ -9,7 +9,7 @@ from typing_extensions import TypedDict
 from useq._channel import Channel  # noqa: TC001  # noqa: TCH001
 from useq._enums import AXES, Axis
 from useq._mda_event import Channel as EventChannel
-from useq._mda_event import MDAEvent, ReadOnlyDict
+from useq._mda_event import MDAEvent, PropertyTuple, ReadOnlyDict
 from useq._utils import _has_axes
 from useq._z import AnyZPlan  # noqa: TC001  # noqa: TCH001
 
@@ -30,7 +30,7 @@ class MDAEventDict(TypedDict, total=False):
     y_pos: float | None
     z_pos: float | None
     sequence: MDASequence | None
-    # properties: list[tuple] | None
+    properties: list[PropertyTuple] | None
     metadata: dict
     reset_event_timer: bool
 
@@ -113,6 +113,7 @@ def _iter_sequence(
     event_kwarg_overrides: MDAEventDict | None = None,
     position_offsets: PositionDict | None = None,
     _last_t_idx: int = -1,
+    _last_p_idx: int = -1,
 ) -> Iterator[MDAEvent]:
     """Helper function for `iter_sequence`.
 
@@ -174,6 +175,11 @@ def _iter_sequence(
         event_kwargs.update(_xyzpos(position, channel, sequence.z_plan, grid, z_pos))
         if position and position.name:
             event_kwargs["pos_name"] = position.name
+        # include position properties only when p-index changes
+        p_idx = index.get(Axis.POSITION, -1)
+        if position and position.properties and p_idx != _last_p_idx:
+            event_kwargs["properties"] = list(position.properties)
+        _last_p_idx = p_idx
         if channel:
             event_kwargs["channel"] = EventChannel.model_construct(
                 config=channel.config, group=channel.group
@@ -223,6 +229,7 @@ def _iter_sequence(
                     event_kwarg_overrides=pos_overrides,
                     position_offsets=_offsets,
                     _last_t_idx=_last_t_idx,
+                    _last_p_idx=_last_p_idx,
                 )
                 continue
             # note that position.sequence may be Falsey even if not None, for example
--- a/src/useq/_position.py
+++ b/src/useq/_position.py
@@ -6,6 +6,7 @@ import numpy as np
 from pydantic import Field, model_validator
 
 from useq._base_model import FrozenModel, MutableModel
+from useq._mda_event import PropertyTuple
 
 if TYPE_CHECKING:
     from matplotlib.axes import Axes
@@ -44,6 +45,7 @@ class PositionBase(MutableModel):
     z: float | None = None
     name: str | None = None
     sequence: Optional["MDASequence"] = None
+    properties: list[PropertyTuple] | None = None
 
     # excluded from serialization
     row: int | None = Field(default=None, exclude=True)
--- a/tests/fixtures/mda.json
+++ b/tests/fixtures/mda.json
@@ -62,6 +62,7 @@
   "stage_positions": [
     {
       "name": null,
+      "properties": null,
       "sequence": null,
       "x": 10.0,
       "y": 20.0,
@@ -69,6 +70,7 @@
     },
     {
       "name": "test_name",
+      "properties": null,
       "sequence": {
         "autofocus_plan": null,
         "axis_order": [
@@ -103,6 +105,7 @@
           "step": 1.0
         }
       },
+      "properties": null,
       "x": 10.0,
       "y": 20.0,
       "z": 50.0


tlambert03:
https://github.com/pymmcore-plus/pymmcore-plus/pull/593 is the matching PR on pymmcore-plus side

ieivanov:
I think I'd be OK with that if you prefer a minimally invasive solution now. If we chose to go this route, I'll take a closer look to double check that I'm happy with everything. 

I also drafted an alternative, somewhat more invasive implementation. The key is adding this, while aiming to preserve backward compatibility.

```python
class StagePosition(UseqModel):
    """A position for a single stage axis.

    Attributes
    ----------
    pos : float
        Position value.
    stage : str | None
        Device name. Optional for spatial axes (x, y, z), where it overrides
        the default device. For other stages, this is the device name.
    """

    pos: float
    stage: str | None = None


class MDAEvent(UseqModel):
    """
    ...
    position : dict[str, StagePosition]
        A mapping of axis name to StagePosition. For spatial axes ("x", "y", "z"),
        the stage device name is optional and overrides the default device. For
        backward compatibility, x_pos, y_pos, and z_pos can still be used as
        constructor arguments and are available as properties.
    """

    ...
    position: dict[str, StagePosition] = Field(default_factory=dict)
```

I think the axis name in `MDAEvent.position` could also help with https://github.com/pymmcore-plus/useq-schema/issues/196. Here are some more implementation details:

---
### 1. useq-schema: Add `StagePosition` model

**File:** `src/useq/_mda_event.py`

```python
class StagePosition(UseqModel):
    pos: float
    stage: str | None = None  # device name (optional for XYZ, required for other stages)
```

The dict key identifies the axis — `"x"`, `"y"`, `"z"` for spatial axes, or the device name for other stages. `stage` optionally overrides the default device for XYZ.

### 2. useq-schema: Add `position` field to `MDAEvent`, replace `x_pos`/`y_pos`/`z_pos`

**File:** `src/useq/_mda_event.py`

Add the new field:
```python
position: dict[str, StagePosition] = Field(default_factory=dict)
```

Remove `x_pos`, `y_pos`, `z_pos` as Pydantic fields. Replace with **properties**:
```python
@property
def x_pos(self) -> float | None:
    sp = self.position.get("x")
    return sp.pos if sp else None

@property
def y_pos(self) -> float | None:
    sp = self.position.get("y")
    return sp.pos if sp else None

@property
def z_pos(self) -> float | None:
    sp = self.position.get("z")
    return sp.pos if sp else None
```

Add a **before-mode model validator** that converts legacy `x_pos`/`y_pos`/`z_pos` kwargs into `position` entries:
```python
@model_validator(mode="before")
@classmethod
def _validate_positions(cls, values):
    if isinstance(values, dict):
        position = dict(values.pop("position", {}))
        for axis in ("x", "y", "z"):
            key = f"{axis}_pos"
            if key in values:
                val = values.pop(key)
                if val is not None:
                    if isinstance(val, (int, float)):
                        position[axis] = {"pos": float(val)}
                    else:
                        position[axis] = val
        if position:
            values["position"] = position
    return values
```

Add a **model validator** that checks x and y have the same stage (when both are set):
```python
@model_validator(mode="after")
def _validate_xy_stage(self):
    x = self.position.get("x")
    y = self.position.get("y")
    if x and y and x.stage != y.stage:
        raise ValueError("x and y positions must use the same XY stage device")
    return self
```

**Serialization**: A model serializer outputs the compact legacy format when possible:
- If all entries are simple XYZ with no stage names → serialize as legacy `x_pos: float, y_pos: float, z_pos: float` keys
- Otherwise → serialize as `position: {...}`

The old `_sx`, `_sy`, `_sz` field serializers are removed.

`MDAEvent.Kwargs` TypedDict includes both legacy and new forms.

### 3. useq-schema: Update `PositionBase`

**File:** `src/useq/_position.py`

Add fields to `PositionBase` (included in serialization, omitted when unset via `exclude_unset=True`):
```python
xy_stage: str | None = None
z_stage: str | None = None
other_stages: dict[str, float] = Field(default_factory=dict)
```

### 4. useq-schema: Propagate position through event iteration

**File:** `src/useq/_iter_sequence.py`

In `_xyzpos()`: construct `StagePosition` objects carrying `stage` from the Position. Also include `other_stages` entries:
```python
position = {}
xy_stage = getattr(pos, "xy_stage", None)
z_stage = getattr(pos, "z_stage", None)
if x_pos is not None:
    position["x"] = StagePosition(pos=x_pos, stage=xy_stage)
if y_pos is not None:
    position["y"] = StagePosition(pos=y_pos, stage=xy_stage)
if z_pos is not None:
    position["z"] = StagePosition(pos=z_pos, stage=z_stage)
for stage_name, val in getattr(pos, "other_stages", {}).items():
    position[stage_name] = StagePosition(pos=val, stage=stage_name)
return {"position": position}
```

---
I drafted an implementation in https://github.com/pymmcore-plus/useq-schema/pull/267 but I don't think I'm quite happy with it yet. It may still give you an idea of what this may looks like. Happy to iterate more on this or pick a solution and go ahead with it.

tlambert03:
yeah, if it's ok with you, I think I prefer this solution #266 to #267 ... at least for now.

Importantly, I think they both allow us to achieve the same *behavior*, and this one is arguably even more flexible (since it doesn't limit you to setting only stage properties per-position.

Your solutions all definitely read with more immediate model clarity... but they also begin to leak more of the exact details of the MMCore stage device model, which at this point, I'm still trying to avoid.  (I feel particularly uncertain about adding three additional specific device-aware fields to the Position object).  It's also 250+ lines of non-test code vs 14 here.  It all feels a bit more like a deprecation risk.

... do you feel strongly about any of this?  Or would you be happy enough to try out this solution for now, and see where it takes us?

ieivanov:
>... do you feel strongly about any of this? Or would you be happy enough to try out this solution for now, and see where it takes us?

Nope, happy to go with #266, I agree that it still achieves the behavior I need. We can reconsider if we run into oddities as we try to use it. I'll close #267.

We can revisit these models during the migration to v2, I'll read over https://github.com/pymmcore-plus/useq-schema/blob/main/docs/v2-migration.md to get more context. Is there a timeline for v2?

tlambert03:
thanks.

> Is there a timeline for v2?

i don't know when all the downstream pymmcore stuff will convert to it.  but `useq.v2` is already importable in the current release.  but yeah:  I do expect stuff to break if you just try to use it with pymmcore-plus.  the idea in pymmcore-plus would be to just cast to v2.  (v2 can easily convert a v1 version to v2, so people still using v1 would be fine).   
It's not super high on the priority list cause everything otherwise mostly works.  The biggest reason I want to update is that v2 handles the concept of an infinite iterator (something where `.size()` fails) much better.  and that helps with things like `time_plan={"interval": 0, "duration": 10}`.

ieivanov:
Got it, thanks! I'll the v2 documentation in more detail, and I'll let you know if I see strong use cases for us.

PR #268, opened 2026-03-24
feat: add Position plate_row/plate_col and configurable name_pattern to grid plans

ieivanov:
## Summary

Adds well-plate position support to `PositionBase`, enabling HCS zarr output from plain stage positions + grid plan without requiring `WellPlatePlan`.

### New fields on `PositionBase`

- **`plate_row: int | str | None`** — Row for well plate positions. Accepts 0-based int (0 → "A") or string ("A") used as-is.
- **`plate_col: int | str | None`** — Column for well plate positions. Accepts 0-based int (0 → "1") or string ("1") used as-is. In YAML, unquoted numbers are parsed as int, so use quotes for string columns (`plate_col: "1"` vs `plate_col: 1`).

### Name generation and validation

- When `plate_row`/`plate_col` are set, the position name is **always derived** from them (e.g., `plate_row=0, plate_col=0` → `"A1"`).
- Providing an explicit name that matches is accepted; a mismatched name raises `ValueError`.
- `WellPlatePlan.selected_well_positions` and `all_well_positions` now populate `plate_row`/`plate_col`.

### Configurable `name_pattern` on grid plans

- New `name_pattern` field on `_GridPlan` (default: `"{idx:04d}"`).
- Supports `{row}`, `{col}`, `{idx}` format variables.
- Example: `name_pattern: "row_{row:03d}_col_{col:04d}"` → `"row_000_col_0000"`.

### Composite `MDAEvent.pos_name`

- For plate positions with a grid plan, `pos_name` is composed as `"{position_name}_{grid_name}"` (e.g., `"A1_0000"`).
- Non-plate positions with grids are unchanged (`pos_name` = position name only).

### Naming behavior

| Input | Result |
|-------|--------|
| `Position(plate_row=0, plate_col=0)` | name=`"A1"` (auto-generated) |
| `Position(name="A1", plate_row=0, plate_col=0)` | name=`"A1"` (matching, accepted) |
| `Position(name="B9", plate_row=0, plate_col=0)` | `ValueError` |
| `Position(plate_row="A", plate_col="1")` | name=`"A1"` (string pass-through) |
| Plate position + grid | pos_name=`"A1_0000"` (composite) |
| Regular position + grid (no plate) | pos_name=`"MyPos"` (unchanged) |

### YAML example

```yaml
stage_positions:
- x: 1000
  y: 1000
  plate_row: 0
  plate_col: 0
  # name auto-generated as "A1"

grid_plan:
  rows: 2
  columns: 2
  fov_width: 180
  fov_height: 180
  # name_pattern: "row_{row:03d}_col_{col:04d}"  # optional
```

## Backward compatibility

All changes are backward compatible:

| Change | Why it's safe |
|--------|---------------|
| `plate_row`/`plate_col` on `PositionBase` | New optional fields, default `None`. Old data deserializes fine. |
| `name_pattern` on `_GridPlan` | Default `"{idx:04d}"` produces identical output to the old `f"{str(idx).zfill(4)}"`. |
| Name auto-generation | Only fires when `plate_row`/`plate_col` are set — never for existing positions. |
| Composite `pos_name` | Only fires when `position.plate_row is not None`. |
| `WellPlatePlan` positions gain `plate_row`/`plate_col` | Additive — existing code reading `x`, `y`, `name` is unaffected. |

**Note**: Serialization output now includes `plate_row`, `plate_col`, and `name_pattern` as new fields. Deserialization of old data is unaffected.

Companion PR: pymmcore-plus/ome-writers#124

## Test plan
- [x] All 385 tests pass (360 existing + 25 new)
- [x] Name auto-generation from int and str plate_row/plate_col
- [x] Mismatched name validation (ValueError)
- [x] JSON and YAML serialization round-trips
- [x] Propagation through Position.__add__
- [x] WellPlatePlan populates plate_row/plate_col
- [x] Grid name_pattern (default, custom, serialization)
- [x] Composite MDAEvent.pos_name for plate positions with grid

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/useq/_grid.py
+++ b/src/useq/_grid.py
@@ -524,7 +524,10 @@ class GridFromPolygon(_GridPlan[AbsolutePosition]):
             pos = []
         for idx, (x, y, r, c) in enumerate(pos):
             yield AbsolutePosition(
-                x=x, y=y, row=r, col=c,
+                x=x,
+                y=y,
+                row=r,
+                col=c,
                 name=self.name_pattern.format(row=r, col=c, idx=idx),
             )
 
--- a/src/useq/_grid.py
+++ b/src/useq/_grid.py
@@ -61,6 +61,9 @@ class _GridPlan(_MultiPointPlan[PositionT]):
         Height of the field of view in microns. If not provided, acquisition engines
         should use current height of the FOV based on the current objective and camera.
         Engines MAY override this even if provided.
+    name_pattern : str
+        Format pattern for grid position names. Supported variables are
+        ``{row}``, ``{col}``, and ``{idx}``. By default, ``"{idx:04d}"``.
     """
 
     overlap: tuple[float, float] = Field(default=(0.0, 0.0), frozen=True)
--- a/src/useq/_grid.py
+++ b/src/useq/_grid.py
@@ -65,6 +65,12 @@ class _GridPlan(_MultiPointPlan[PositionT]):
 
     overlap: tuple[float, float] = Field(default=(0.0, 0.0), frozen=True)
     mode: OrderMode = Field(default=OrderMode.row_wise_snake, frozen=True)
+    name_pattern: str = Field(
+        default="{idx:04d}",
+        frozen=True,
+        description="Format pattern for grid position names. "
+        "Supported variables: {row}, {col}, {idx}.",
+    )
 
     @field_validator("overlap", mode="before")
     @classmethod
@@ -137,7 +143,7 @@ class _GridPlan(_MultiPointPlan[PositionT]):
                 y=y0 - r * dy,
                 row=r,
                 col=c,
-                name=f"{str(idx).zfill(4)}",
+                name=self.name_pattern.format(row=r, col=c, idx=idx),
             )
 
     def __iter__(self) -> Iterator[PositionT]:  # type: ignore [override]
@@ -517,7 +523,10 @@ class GridFromPolygon(_GridPlan[AbsolutePosition]):
         except ValueError:
             pos = []
         for idx, (x, y, r, c) in enumerate(pos):
-            yield AbsolutePosition(x=x, y=y, row=r, col=c, name=f"{str(idx).zfill(4)}")
+            yield AbsolutePosition(
+                x=x, y=y, row=r, col=c,
+                name=self.name_pattern.format(row=r, col=c, idx=idx),
+            )
 
     def _cached_tiles(
         self,
--- a/src/useq/_grid.py
+++ b/src/useq/_grid.py
@@ -63,7 +63,7 @@ class _GridPlan(_MultiPointPlan[PositionT]):
         Engines MAY override this even if provided.
     name_pattern : str
         Format pattern for grid position names. Supported variables are
-        ``{row}``, ``{col}``, and ``{idx}``. By default, ``"{idx:04d}"``.
+        `{row}`, `{col}`, and `{idx}`. By default, `"{idx:04d}"`.
     """
 
     overlap: tuple[float, float] = Field(default=(0.0, 0.0), frozen=True)
--- a/src/useq/_position.py
+++ b/src/useq/_position.py
@@ -121,9 +121,7 @@ class PositionBase(MutableModel):
         """
         if self.plate_row is not None and self.plate_col is not None:
             if isinstance(self.plate_row, int) and isinstance(self.plate_col, int):
-                well_name = (
-                    f"{_index_to_row_name(self.plate_row)}{self.plate_col + 1}"
-                )
+                well_name = f"{_index_to_row_name(self.plate_row)}{self.plate_col + 1}"
                 if self.name is not None and self.name != well_name:
                     raise ValueError(
                         f"Position name {self.name!r} does not match "
--- a/src/useq/_position.py
+++ b/src/useq/_position.py
@@ -46,6 +46,8 @@ class PositionBase(MutableModel):
     name: str | None = None
     sequence: Optional["MDASequence"] = None
     properties: list[PropertyTuple] | None = None
+    plate_row: int | None = None
+    plate_col: int | None = None
 
     # excluded from serialization
     row: int | None = Field(default=None, exclude=True)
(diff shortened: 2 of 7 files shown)


ieivanov:
I think I'm happy with this PR, happy to discuss more!

tlambert03:
> without requiring WellPlatePlan

just so that I have the full picture ... can you elaborate on why you *don't* want to use WellPlatePlan for describing positions in a well plate?

ieivanov:
>just so that I have the full picture ... can you elaborate on why you don't want to use WellPlatePlan for describing positions in a well plate?

Ah, yes, sorry that wasn't clear. `WellPlatePlan` is great when you want to construct a set of positions with regular spacing - preset distance between wells. When we image well plates, we often tune the Z position for each well (we use high NA objectives what can be sensitive to small warps in the well bottom) and may occasionally tune the well center to account for uneven seeding of the cells. This is fairly manual, but worthwhile for some acquisitions.

At a higher level, while a `WellPlatePlan` is a `Sequence[Position]`, not every `Sequence[Position]` can be a `WellPlatePlan`. I'd like `WellPlatePlan` to be convenience method for constructing a set of positions in a well plate, rather than a requirement. Currently, in `ome-writers` we require a `useq.WellPlatePlan` to write HCS zarr stores since `useq.Position`s themselves don't carry enough information to construct a well plate.

```
def _plate_from_useq(seq: useq.MDASequence) -> Plate | None:
    """Convert a useq WellPlatePlan to an ome-writers Plate."""
    import useq

    useq_plate = seq.stage_positions
    if not isinstance(useq_plate, useq.WellPlatePlan):
        return None
```

This feature is added in https://github.com/pymmcore-plus/ome-writers/pull/124

tlambert03:
> preset distance between wells

ah, perfect.  Yes, makes sense. thank you

tlambert03:
> I'd like WellPlatePlan to be convenience method for constructing a set of positions in a well plate, rather than a requirement.

i don't think I want to go in that direction (making `WellPlatePlan` a literal convenience method) ... since that loses serializability.  I think having it *be* a model, that also happens to be a `Sequence[Position]` is the correct model here.  but, I do agree that it lacks the flexibility for this case, and supporting any `sequence[Position]` is better.  So, generally agree with the overall point here

ieivanov:
>I think having it be a model, that also happens to be a Sequence[Position] is the correct model here.

Yes, I agree, I misspoke there

ieivanov:
I'm also contemplating https://github.com/ieivanov/useq-schema/pull/1, I'd be happy with or without it

src/useq/_grid.py
```diff
@@ -61,10 +61,19 @@ class _GridPlan(_MultiPointPlan[PositionT]):
         Height of the field of view in microns. If not provided, acquisition engines
         should use current height of the FOV based on the current objective and camera.
         Engines MAY override this even if provided.
+    name_pattern : str
+        Format pattern for grid position names. Supported variables are
+        ``{row}``, ``{col}``, and ``{idx}``. By default, ``"{idx:04d}"``.
```
tlambert03:
```suggestion
        `{row}`, `{col}`, and `{idx}`. By default, `"{idx:04d}"`.
```
(double backticks are rst format but we use mkdocs/zensical everywhere, no sphinx)

src/useq/_position.py
```diff
@@ -46,6 +69,8 @@ class PositionBase(MutableModel):
     name: str | None = None
     sequence: Optional["MDASequence"] = None
     properties: list[PropertyTuple] | None = None
+    plate_row: int | str | None = None
+    plate_col: int | str | None = None
```
tlambert03, line 74: I'm a bit uncertain about `int | str` here.  I see in your tests that you've used this for things like `Position(plate_row="fish0", plate_col="neuromast0")`... but that feels like it's overloading one concept (which row/col) to do double duty to store metadata that should arguably go elsewhere (either in the position name itself (which is already a totally open string) or in the sequence metadata somewhere (where you could map your row/col indices to treatments/metadata).  Also, thinking ahead to how this information would be used to write an OME-Zarr, we *have* to know the row/col index (https://imaging-formats.github.io/yaozarrs/API_Reference/yaozarrs.v05/?h=row#yaozarrs.v05._plate.PlateWell) ... so allowing the user to provide a string here means we would have difficulty guaranteeing the ability to write ome-zarr.

I'm inclined to reduce this back to just `int` for now.

src/useq/_position.py
```diff
@@ -85,6 +110,34 @@ def __round__(self, ndigits: "SupportsIndex | None" = None) -> "Self":
         # not sure why these Self types are not working
         return type(self).model_construct(**kwargs)  # type: ignore [return-value]
 
+    @model_validator(mode="after")
+    def _name_from_plate(self) -> "Self":
```
tlambert03, line 131: this all feels like application-level logic to me.  I don't think we should auto-calculate names from other fields.  Just store the data, and leave this sort of concern (autogeneration of naming) to whoever is *using* the data, like a writer, or a GUI.

tlambert03 (commented):
@ieivanov 

After thinking about this a bit, I'm definitely supportive of adding 

```python
class PositionBase:
    plate_row: int | None = None
    plate_col: int | None = None
```

but I think most of the rest of this PR (naming related stuff) is too opinionated for this level.  I think we should try to just store the data and leave the name autogeneration to downstream applications that might have different opinions on how it should be done.

Would you be ok if I heavily reduced this PR to just adding the plate_row/plate_col fields?

tlambert03:
@ieivanov, I merged #269, which branched off of and simplified this PR to just add the plate_row/plate_col stuff, and I've merged main back into your branch here.  So, this PR is now just about adding naming conveniences and formatting.  This is all stuff that I'm not entirely sure about... but we can continue the discussion here when you like.  I think the primary need for the purpose of ome-writers (and persistent plate metadata without using a WellPlate object) is accomplished by #269

ieivanov:
Thanks for thinking through this! I agree with you that naming logic is better suited to `ome-writers`. I think #269 accomplished what I needed from `useq-schema`.

There is a bug in the current implementation of naming in `ome-writers`, however. If I try an acquisition with 

```yaml
stage_positions:
- x: 14000
  y: 11000
  plate_row: 0
  plate_col: 0
- x: 23500
  y: 11000
  plate_row: 0
  plate_col: 1

grid_plan:
  rows: 2
  columns: 2
  fov_width: 180
  fov_height: 180
  overlap: -10
```

I get an error due to duplicate names in the zarr store. I'll follow up with an issue on that repo.

